PR #1, opened 2018-09-10
Add infrastructure to enable additional references in documents

197g:
Rough sketch, but provides the enough code to enable additional inline link magic:

```
![](//TOC)
```

Glue code to actually add this to the latex provider will be added if the proposed syntax is accepted and when I have become more familiar with the different traits.

The basic consideration is that urls provide, with domains, a terse syntax to provide what we would expect from inline (image) links but can at the same time be reused to implement other, non-image link types without loss of expressivity. Conceptually, we treat documents as residing under an url with origin `pundoc://document/` where the root corresponds to the project. Other domain names can then be used for extended functionality (using shorthand `//new-domain` for changing domain but not scheme) while both standard local files (which begin with at most `/`) and linked http-available files (which require `http[s]://` for access) remain consistent in their syntax.

Additional benefits: It will be impossible by default to refer to documents outside the current workspace, such as `/etc/passwd` or any other unwated file. This should later be extended by options to selectively allow it.


src/resolve/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+pub struct Source {
+    inner: InnerSource,
+}
+
+#[derive(Clone, Eq, PartialEq, Hash)]
+enum InnerSource {
+    /// Implementation detail with vetted access.
+    ///
+    /// The burden is on this implemention to ensure that no confused deputy situation arises. This
+    /// implements special such as `//TOC`. The `//document` host allows qccess to the all project
```
oberien: `qccess` -> `access`

src/resolve/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// Implemented with the `file` scheme.
+    Local(Url),
+
+    /// Any other arbitrary url.
+    Remote(Url),
+}
+
+enum Resource {
+    File(PathBuf),
+    Abstract(Box<io::Read>),
```
oberien: Please use `dyn Read` syntax.

src/resolve/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /// Add a standard provider for a document tree.
+    pub fn add_documents<P: Into<PathBuf>>(&mut self, base: P) {
+        let documents = Documents(base.into());
+        self.add_provider("document", Box::new(documents));
+    }
+
+    pub fn add_provider<H: Into<String>>(&mut self, host: H, provider: Box<DocumentProvider>) {
+        let host = Host::Domain(host.into());
```
oberien, line 79: You could use a `Cow<'static, str>` here to avoid additional allocations, but because this is only executed during startup of pundoc, it doesn't really matter.
Nvm, `url::Host::parse` is only implemented for `Host<String>` anyways.

src/resolve/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+        self.as_url()
+            .join(reference)
+            .map(Self::from_url)
+    }
+
+    fn as_url(&self) -> &Url {
+        match self.inner {
+            InnerSource::Implementation(ref url) => url,
+            InnerSource::Local(ref url) => url,
+            InnerSource::Remote(ref url) => url,
```
oberien: match_default_bindings allows you to leave out the `ref`s when matching on `&self.inner` instead of `self.inner`.

src/resolve/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+            _ => None,
+        }
+    }
+
+    /// Turn this document in a readable data stream.
+    ///
+    /// This will not give a sensible result for all documents. Some documents might be filled
+    /// during the processing of the markdown document. These will of course initially appear empty
+    /// or contain data of previous runs.
+    pub fn into_reader(self) -> io::Result<Box<io::Read>> {
```
oberien: `dyn Trait`

oberien (changes requested):
Sorry for the late response.

I like this approach in general, but I think this is somewhat overcomplicated and doesn't really fit within the structure of how pundoc performs translations.

The general idea of pundoc's generation is that the generator reads the Event stream as produced by pulldown-cmark. It then forwards the events to the respective defined code generator. The codegen is defined by a trait with an associated type for each event. The generator keeps an explicit stack of all current generators (e.g. list → list → inline-code → text), allowing the code generators to access the stack (e.g. for `\begin{enumerate}` to find out the nesting to set the corresponding `\enumiii` variable if the list starts from a higher number).

The generator intercepts some events to test for special commands. These special commands should again be forwarded to the respective codegen type. (At a later point I was thinking about making interceptors a trait, but for now plain methods on the generator should be good enough.)

Thus my plan for glue-code for this URL-PR is to find "abnormal" links / includes in the generator by intercepting `Event::Start(Tag::Link)` / `Event::Start(Tag::Image)` and testing if they are commands / URLs which need to be handled differently. If that's the case, the respective codegen is called to generate code in its language (latex / html / …).

Thus my idea would be for the resolver to just take a link / uri in its `resolve` method and return the respective event / information to the generator such that the generator can call the respective codegen.  
The resolver should be more or less a black-box taking a URL and returning information needed for the generator to call the corresponding codegen.

We should definitely have a cache of files downloaded via URIs to allow working without internet. We should autodetect that internet is missing and fall back to the cache, but also have command line options to both only use the cache and not use the cache (fetch freshly from the internet).

Regarding your code I have to admit that I don't fully understand which design you're going for, because I'm missing the usage you had in mind. That's why for now I have just commented the code generally.

Also a thing to consider with your code is that it appears to be missing support for DocumentBuilders like pdf-includes. Also markdown-includes might be more complicated and not fit into your current structure. Their current implementation is [here](https://github.com/oberien/pundoc/blob/master/src/gen/generator.rs#L104).

oberien:
Now let's get back to the general discussion around includes and URIs / URLs that started [here](https://github.com/oberien/pundoc/commit/4b22579f68b13f14c9f4664d14bd9ef3de5dc1d4#r30417238) and [here](https://github.com/oberien/pundoc/commit/4b22579f68b13f14c9f4664d14bd9ef3de5dc1d4#r30416976). At first I wasn't really convinced by using URLs for includes. Having looked at this PR I start to like them more, because they add a general common syntax for includes of any kind. I still have 3 problems with them currently though:

1. URLs seem very nice and amazingly complex to be nicely usable by advanced users, but I think that it might be too complicated for people that just want to use pundoc and don't have an informatics degree to understand URIs and their usage / implications. It might feel natural for people knowing about URIs to write ``, but it will probably feel very weird for others. Also a table of contents isn't actually an include but rather a summary of already existing things within the file.
2. The syntax feels unnatural and not like proper markdown, just like my earlier proposal of `!!foo{bar}`. I still like `[TOC]` and `[include foo.bar]` more, because they are obvious and feel natural to markdown. The `alt-text` syntax is more like an "embed this please" than an "include and render this as well please" IMO.
Let's take a look at how others handle includes:
    * [markdown-include (js)](https://github.com/sethen/markdown-include): C-style includes `#include "markdown-file.md"`. I definitely don't like that :)
    * [markdown-include (python)](https://github.com/cmacmackin/markdown-include): `{!file.md!}`. I also don't like that syntax. Feels more like latex than markdown.
    * [pandoc-include](http://hackage.haskell.org/package/pandoc-include):
        ````
        ```include
        path/to/file.md
        ```
        ````
        This looks better to me, but still not amazing, because it adds more boilerplate than I'd like.
    * [MarkdownPP](https://github.com/jreese/markdown-pp): `!INCLUDE "chapters/preface.mdpp"`. Same as with C-style includes.
    * [cat.pl](https://github.com/district10/cat/blob/master/bin/cat.pl): `@include <-=path=`. This is super weird IMHO.
    * Marked 2: `<<[chapters/chapter1.md]`. I really like that syntax, but it still doesn't feel fully intuitive to me.
    * [This commonmark thread](https://talk.commonmark.org/t/embedded-audio-and-video/441) has some other interesting ideas.

    So finally I still don't know what the proper syntax should be. Currently I'm still in favour of `[TOC]` and `[include foo.md]` compared to `` and ``, because they are easier to understand and to remember and someone who doesn't know pundoc can still understand what's meant. Unfortunately I don't know how to incorporate URIs into that.

3. Custom commands that are not includes might need to abuse include-URIs or need a new syntax anyways, so why shouldn't we have a common syntax in the first place? Right now I can't think of any custom command that's not an include, though.

197g:
The current state of the approach is simply to offer a resolution framework that maps urls to file paths/Files/io::Read. I tried to forward the decision what to actually do with that information to the pundoc backend. The focus was to make sure that resolution of references can happen in the same manner across all document classes. I'll try to demonstrate usage in the latex backend.

> Thus my idea would be for the resolver to just take a link / uri in its resolve method and return the respective event / information to the generator such that the generator can call the respective codegen.
The resolver should be more or less a black-box taking a URL and returning information needed for the generator to call the corresponding codegen.

That was basically the idea here although the implementation was probably vague. With the additional thought that if you recursively include documents from the internet, those should preferably reference their own source for relative links and not be able to resolve local files (or only after specifically approving that), thus adding a second `Source` argument (bikeshedding issue here, I don't like the name).

With the additional comments on how exactly the current translation works, the interface offered by document likely needs some work.

> We should definitely have a cache of files downloaded via URIs to allow working without internet. We should autodetect that internet is missing and fall back to the cache, but also have command line options to both only use the cache and not use the cache (fetch freshly from the internet).

Or just adhere to the cache directives given in the https request. Include optional etags in the cache as well and we should be good to go. This would be in the spirit of the current cargo approach, which only provides an `offline` option and is online by default.

> Also a thing to consider with your code is that it appears to be missing support for DocumentBuilders like pdf-includes. Also markdown-includes might be more complicated and not fit into your current structure. Their current implementation is here.

I'm not sure I understand what you mean by `DocumentBuilders like pdf-includes` but markdown includes should be done by checking the type of the document. Oops, I missed adding a `pub fn mime_hint() -> Option<MimeType>` to the interface of `Document`.

> URLs seem very nice and amazingly complex to be nicely usable by advanced users, but I think that it might be too complicated for people that just want to use pundoc and don't have an informatics degree to understand URIs and their usage / implications.

Google seems to think the same ;) But we don't necessarily have to explain it as full URLs. I would explain `` as `embed this external document here`, as opposed to merely link it. Then `//` could simply be introduced as a special reference to documents provided by the pundoc processor. And only later in a guide would one explain the synthesis of both being valid urls from the current base. At least for me, that is more natural to write than extending markdown with another embedder syntax and a completely different resolution mechanism.
 
> The syntax feels unnatural and not like proper markdown, 

The common mark thread is interesting, everything else looks even worse than both the current and my approach. Especially the proposed alternatives on how to add auxiliary information. From my perspective, the inline link syntax has the advantage of being able to specify additional information to the processor in the `alt` section. For example, if the file type can not be deduced from the standard algorithms (which is response header then file extension if memory serves me right). But some would suggest this qualifies as abuse.

Edit: Of course, the design with the alt section would need to consider that there might be content in it as well. But at first one could merely use it for media type anyways, as a suffix or prefix. Then we could just strip it once. For example:

```
![An audio file with horse whinnying; audio/mp4](/horses.mp4)
```

> Custom commands that are not includes might need to abuse include-URIs or need a new syntax anyways, so why shouldn't we have a common syntax in the first place? Right now I can't think of any custom command that's not an include, though.

That syntax could be common with inline images as well is what I'm saying. Other custom commands that I could imagine:

* `Plugins`: Those should be specified in the file before usage, imho. But you could do that in a meta section as well and it might be unecessary.
* `Generated files from pipes`: As a basic example I would like to include dot files and svg as images. Of course, for these examples one could still just detect the content type and hard code filters.
* `doi` based automatic citations. Simply paste the doi and the processor tries to resolve the name, title, etc. Not sure if it would work. Same could be enabled for `arxiv`! That's more a matter of being lazy of course but hey, not having to type a lot is the whole point of markdown.

197g:
Additional comment to includes
```
```include
path/to/file.md
```

is the only syntax where specifying additional auxiliary information (`{ size=10 duration=10 cycle=forever }` as discussed in the commonmark thread) would feel natural to me. But it is too verbose for standard includes. On the other hand, specifying it such as in `data:image/png;base64,` is too terse for a plain text file.

oberien:
> The current state of the approach is simply to offer a resolution framework that maps urls to file paths/Files/io::Read. I tried to forward the decision what to actually do with that information to the pundoc backend. The focus was to make sure that resolution of references can happen in the same manner across all document classes. I'll try to demonstrate usage in the latex backend.

I think I understand now. With that information I'll give the code review another try.

> That was basically the idea here although the implementation was probably vague. With the additional thought that if you recursively include documents from the internet, those should preferably reference their own source for relative links and not be able to resolve local files (or only after specifically approving that), thus adding a second `Source` argument (bikeshedding issue here, I don't like the name).

I have an idea how to handle that. We already have an explicit stack of where in the document (i.e. how deep) we are. We could add something like `DocumentInclude(Source)` to the stack as well, allowing us to pass the stack's topmost `source` to the resolver's `resolve` method, which allows us to handle permissions / other includes within the source as done right now.
The very first element on the stack should be `DocumentInclude(inputfile.md)`.
I think that this is the easiest and most future-proof way to do this. Additionally, this allows codegen units to see if they are executed within includes and allow them to render differently based on that.

This means the general approach is this:

1. Upon finding a link in the `Generator`, let the resolver resolve it.
2. The resolver downloads (tries to download) the resource if it's an online link (defaults to the cached file if download failes). Otherwise it resolves to the specified command or local file.
3. If no mime-type could be determined for downloaded / local files, an error is returned (because how should we properly handle includes / embeds if we don't even know their type).
4. The resolver caches the file with its mime-type.
5. The resolver returns the path to the cached file / the command to the `Generator`.
6. The generator adds the appropriate state to the stack.
7. The generator branches on the mime-type and handles it appropriately. If the result was a command, the generator forwards / handles it appropriately.

> Or just adhere to the cache directives given in the https request. Include optional etags in the cache as well and we should be good to go. This would be in the spirit of the current cargo approach, which only provides an `offline` option and is online by default.

I definitely want to be able to work offline without specifying any argument. I personally dislike the design decision cargo went for. We should send out requests and if they don't return something within the first second or so (configurable and might need tweaking), we default to the cached version.

> Oops, I missed adding a `pub fn mime_hint() -> Option<MimeType>` to the interface of `Document`.

I don't think the mime-type should be an option. We can only handle an embed / include if we know it's mime type.

> The common mark thread is interesting, everything else looks even worse than both the current and my approach. Especially the proposed alternatives on how to add auxiliary information. From my perspective, the inline link syntax has the advantage of being able to specify additional information to the processor in the `alt` section. For example, if the file type can not be deduced from the standard algorithms (which is response header then file extension if memory serves me right). But some would suggest this qualifies as abuse.

Instead of the `alt` section, we could also abuse the `title` of links. While it qualifies as abuse to some, I think it's fine to do so.

> […] is the only syntax where specifying additional auxiliary information ({ size=10 duration=10 cycle=forever } as discussed in the commonmark thread) would feel natural to me.

I think we could allow putting the mime-type in the `alt` section while using the `title` as aux infomation.
We need to keep in mind that we also need a way to specify labels (which is what I wanted to use the `title` for. Thus, I think specifying the type and aux info in the `alt` section is best.

oberien:
I think the general structure with the design proposed in the previous comment should be to have a `Resolver::resolve(ctx: &Context, link: String) -> (Include, Option<Context>)` method which is used by the `Generator`.

The types should be something among the lines of this:

```rust
struct Resolver { … }

impl Resolver {
    /// Resolves given link within the context of the given Context
    ///
    /// The first returned value is the include to handle in the `Generator`, the second
    /// return value is the optional context to be stored on the stack.
    /// (Maybe it's better to have the context in the `Include` where it's needed (it's only needed for Markdown includes AFAICT))
    pub fn resolve(context: &Context, link: String) -> (Include, Option<Context>) { … }
}

/// Context passed to and returned from the resolver
///
/// It's put on the stack and is only an implementation detail to the `Generator` for the `Resolver`
struct Context {
    source: Source // or however you want to call it
}

impl Context {
    /// Creates the initial Context of the input markdown file / stream
    ///
    /// The parameter is the folder the input file resides in or the folder pundoc was executed from if the input source is stdin
    pub fn from_root<P: AsRef<Path>>(path: P) -> { … }
}

/// Include return by `Resolver::resolve`
///
/// This is matched on by the `Generator` to call the respective appropriate handler
enum Include {
    Command(Command),
    Markdown(PathBuf),
    Image(PathBuf, Option<ImageMeta>),
    Pdf(PathBuf, Option<PdfMeta>),
    …
}

struct ImageMeta {
    width: Option<String>,
    height: Option<String>,
    …
}

enum Command {
    Toc,
    …
}
```

This way the mime-type is translated within the resolver to predefined output formats / commands.

197g:
> I don't think the mime-type should be an option. We can only handle an embed / include if we know it's mime type.

It could still be an option to suggest that the resolver might not be able to determine the correct type. But `Result<MimeType, ()>` is also fine.

> Thus, I think specifying the type and aux info in the alt section is best.

Not only helping humans with disabilities but also our programs, an effective and appropriate use of that section ;)

oberien:
Regarding the syntax, I think we could have both your proposed general URI include syntax and a subset of it with the link syntax. We could translate some special links (like `[TOC]`) to their respective URIs (`"//TOC"`) to allow using both but only have a single resolver implementation.

197g:
New interface provides `Document` as the proposed tuple source and include, providing direct references to both as well as a convenience method for getting the path if desired. I would add the deduction of mime type, and local caching, to the methods in `DocumentBuilder` when the need actually arises. That would have the additional advantage that we can pass a reference to a temporary cache directory from `Resolver` and have some configuration there instead of in a per-`DocumentProvider` basis.

src/resolve/document.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/// Typed representation of the resolved resource.
+///
+/// This is matched on by the `Generator` to call the respective appropriate handler
+#[derive(Debug, PartialEq, Eq)]
+pub enum Include {
+    Command(Command),
+    Markdown(PathBuf),
+    Image(PathBuf, Option<ImageMeta>),
```
oberien: Thinking about it, if everything within `ImageMeta` is already optional, the `ImageMeta` itself should always be provided.

src/resolve/document.rs
```diff
@@ -0,0 +1,105 @@
...
+    Pdf(PathBuf, Option<PdfMeta>),
```
oberien: ~

oberien:
Does this mean that you are finished with this PR? Or are you going to add glue-code as well? Or should I do that?

197g:
PR is a complete unit as is so I think we should merge, at least so we can work on the code on parallel. I'd add interfaces/more constructors for `DocumentBuilder` first, you can do the glue-code.

oberien (approved):

PR #2, opened 2018-09-14
Refactor Resolve

oberien:


src/resolve/providers.rs
```diff
@@ -0,0 +1,109 @@
+use std::io;
+use std::path::{PathBuf, Path, Component};
+
+use url::{Url, Host};
+
+use resolve::{Resource, ResourceBuilder, Source, InnerSource, Include, Command};
+
+// We Java now
```
197g: Rename `ResourceProvider => Server`, `ResourceProviderProvider => ServerProvider`

oberien:
@HeroicKatora Do you plan on reviewing this or can I merge it without your evaluation? (If you'd like to but don't have time right now that's fine, there's plenty of other stuff that needs to be done :-) )

197g:
I'll have a more thorough look on the weekend if you want but the most important aspects of resolution should have been covered by the simple test anyways. Moving the host based resolution into its own type and `impl` and removing the indirection through the hashmap is likely better for now until there is a use case for changing that dynamically.

src/resolve/mod.rs
```diff
@@ -257,8 +196,8 @@ mod tests {
     fn domain_resolves() {
```
197g, line 146: This test needs to be reworked. 

https://github.com/oberien/pundoc/blob/d193191e5178075734fd88d88ff3246fb43dc717/src/resolve/mod.rs#L197-L203

Remove the `struct` and `add_provider` call then the builtin hardcoded mechanism should take over.

197g: Without regards to the review content, is it possible to tag multiple and existing lines in a review? That would have made the comment more helpful.

oberien: Not for a review comment directly, but within the review comment you can use [this](https://github.community/t5/How-to-use-Git-and-GitHub/PR-multiline-review/m-p/6779#M2261).

oberien:
@HeroicKatora I went back to the roots and tried to find out the question we are trying to solve with this PR. Landing at #3 I completely refactored this code again to make it simpler and more straightforward, less overengineered.
Could you please give this version another review?

Issue #3, opened 2018-09-23
Embed all the things


oberien:
Embed all the things: `![alt-text](URL "title")`

* Infrastructure to resolve URL
* transform URL into Include
* handle all URLs
    * care only about url
    * `pundoc://document/` is the default URL resolution base
        * all relative URLs are resolved relative to this
* Sources (based on (schema, domain)-pair)
    * Implementation (`//TOC`, `pundoc://REFERENCES`)
        * "invoke" a codegen function, something implemented
    * LocalRelative (`foo.md`, `/foo.md`, `pundoc://document/foo.md`, `file:///path/to/workdir/foo.md`)
        * local file, relative to latest included markdown file 
        * within context directory
        * can only be done if contex == LocalRelative
    * LocalAbsolute (`file:///etc/passwd`)
        * local file
        * outside of context directory
        * require special permissions from `pundoc`-invoker
            * cli-args with allowed directories
            * ask user on stdin
        * ignored for the time being, maybe added later
    * Remote file (`![](https://placehold.it/100x100)`)
        * downloaded and cached locally, may be served from cache (e.g. if no internet)
        * remote can only include other remote (any??? same domain??? only relative without backtracking??? TODO)
* Resolved to "include"
    * Command (for generator to call codgen to generate according thing):
        * TOC
        * References / Bibliography
        * … potential other commands in the future
    * Images
        * PNG
        * … potential other formats in the future
    * PDF
    * Markdown rendered
        * handle permissions for further includes based on source
    * … potential other results in the future
* Use hash as filename in cache
* MUST disallow local includes from cache folder
    * Attack scenario (remote for obscurity):
        1. I want to build pdf of someone else
        2. I download their repo
        3. They use html backend
        4. Their markdown file includes remote markdown file
        5. Remote wants to acces local file → Disallowed
        6. Skip that include, continue codegen
        7. Their original file now includes file from cache
        8. Allowed, because it's in a subfolder of the project
        9. That file includes my credit card information (technically LocalRelative still can't include anything, but it's already a privilege escalation)
        10. Rest gets generated
        11. Inline JS gets the credit card information from the document and post it to attacker's server

"User-friendly" syntax

* `[TOC]`
* `[include "foo.md"]`
* `[REFERENCES]`
* `![](https://placehold.it/100x100)`

197g:
[...]

PR #4, opened 2018-10-22
Implements downloading of included files

197g:


oberien:
First I saw the diff of 1.3k and though: WTF how and why?!? Then I saw that 1.1k of that is just Cargo.lock ;)

My general idea of downloading stuff from the internet is to have a cache folder in a subfolder of the workdir like `.cache` (which is why I only passed the workdir to the Resolver). Then we use a modified version of servo's http_cache (<https://github.com/servo/servo/blob/910cc23a6e85cced43905f7615065b23bdb54b42/components/net/http_cache.rs>) to find out how long something can be cached / when it needs to be reloaded. In offline mode we can then just *not* redownload even if the cache would suggest otherwise.

Includes passed to the resolver, MUST NOT be from that directory as discussed in #3 .

EDIT: We may be able to avoid the "must not include from cache" problem by generating a random salt on cache creation and including that in hashes of the URL. That should prevent predictable paths to the cache but still allow reproducible hashes as the salt will be saved in the cache dir.

197g:
> Includes passed to the resolver, MUST NOT be from that directory as discussed in #3 .

Since it uses the temp dir, the path name can not be currently guessed anyways. Can we do caching as a separate feature for another pull request? A good suggestion to reuse servo's implementation.

oberien:
> Since it uses the temp dir, the path name can not be currently guessed anyways.

For now yes. Have you seen my idea in the EDIT?

> Can we do caching as a separate feature for another pull request?

Yes, makes for smaller PRs and easier reviews ;) I'll review this tomorrow.

> A good suggestion to reuse servo's implementation.

We'll probably only be able to reuse parts of it, because it uses its own http-client and iirc is only an in-memory and not file-based caching. So we may be able to reuse the caching / cache resolving parts of it, but probably end up rewriting the whole backend. This can also be a different PR after the one implementing cache-directories and access prevention.

197g:
> We'll probably only be able to reuse parts of it, because it uses its own http-client and iirc is only an in-memory and not file-based caching. So we may be able to reuse the caching / cache resolving parts of it, but probably end up rewriting the whole backend.

I would mostly use it as guideline to avoid having to re-evaluate whether our implementation adheres to the referenced RFCs.

> Have you seen my idea in the EDIT?

I'm not sure if hashes are the perfect solution. Unpredictability is only a part of the problem for the download directory. We might later want to have additional auxiliary files for a project and you should also not be able to reference the file containing the salt. Also, file names should be able to contain an arbitrary file extension to avoid compatibility issues with programs expecting specific ones (such as `pdflatex`), this pr already includes a simple renaming that was necessary for the specific `.png` image include. In short, disallowing the folder seems simpler and more general to me.

src/config/mod.rs
```diff
@@ -185,6 +186,7 @@ impl Config {
         Config {
             output,
             out_dir: args.out_dir.unwrap_or(tempdir.path().to_owned()),
+            temp_dir: tempdir.path().to_owned(),
```
oberien, line 194: For now this is an acceptable escape hatch for permission checking with regards to includes from the cache folder, but I opened #10 to remove it once #6 is implemented.

src/resolve/mod.rs
```diff
@@ -111,6 +115,40 @@ mod tests {
...
+ right: `{:?}`"#, left_val, stringify!($right)
+                           format_args!($($arg)+))
+            }
+        }
+    });
+    ($left:expr, $right:pat, $($arg:tt)+) => ({
+        assert_match!($left, $right if true, $($arg)+)
+    });
+}
+
```
oberien, line 134: Did you really just introduce a 30 lines macro to replace 6 additional lines of code? ;)

(also yes, oops, I forgot that `path()` is used in the tests when I removed it when I hit the unused warning. Maybe we should reintroduce that function instead of this macro)

src/resolve/remote.rs
```diff
@@ -0,0 +1,141 @@
+use std::fs::{File, OpenOptions, create_dir_all, remove_file};
```
oberien: Please use `fs::create_dir_all` over `create_dir_all`. Same with `remove_file`.

src/resolve/remote.rs
```diff
@@ -0,0 +1,141 @@
...
+            .send()
+            .and_then(|response| response.error_for_status())?;
+
+        let path = self.target_path(&url);
+
+        create_dir_all(path.parent().unwrap())?;
+
+        // Replace whatever file already existed.
+        // TODO: proper caching
+        let _ = remove_file(&path);
```
oberien: Removing the file shouldn't be necessary as `File::create` will truncate existing files / create new ones already.

src/resolve/remote.rs
```diff
@@ -0,0 +1,141 @@
...
+        let mut file = OpenOptions::new()
+            .write(true)
+            .create_new(true)
+            .open(&path)?;
```
oberien, line 83: `remove_file` followed by this `OpenOptions` is exactly what `File::create` does:

https://github.com/rust-lang/rust/blob/945372d26818f93d6f5cded7b751749e280b67bf/src/libstd/fs.rs#L388-L390

src/resolve/remote.rs
```diff
@@ -0,0 +1,141 @@
...
+
+        io::copy(&mut response, &mut file)?;
```
oberien, line 85: :+1:

src/resolve/remote.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    fn content_type(&self, response: &Response) -> Option<ContentType> {
+        // Get a value content type header if any.
+        let header = response.headers().get(header::CONTENT_TYPE);
+        let header = header.and_then(|value| value.to_str().ok());
+
+        match header {
+            Some("text/markdown") => Some(ContentType::Markdown),
+            Some("image/png") | Some("image/jpeg") => Some(ContentType::Image),
+            Some("application/pdf") => Some(ContentType::Pdf),
```
oberien: Can we use the `mime` crate here?

src/resolve/remote.rs
```diff
@@ -0,0 +1,141 @@
...
+        // http(s) must not be cannot-be-base
+        //
+        // Also, '+' is a reserved character that can not appear unescaped.
+        let path = url.path().replace('/', "+");
+        let path = Path::new(&path);
+
+        // Since pdflatex is picky with file extensions, replace all dots.
+        let stem = path.file_stem()
+            .map(|osstr| osstr.to_str().unwrap())
+            .unwrap_or("no_name")
```
oberien: What if there are multiple no_name files? The last one will be included into the pdf at all places (and possibly be invalid if one is a pdf and the other an image).

src/resolve/remote.rs
```diff
@@ -0,0 +1,141 @@
...
+    fn content_type(&self, response: &Response) -> Option<ContentType> {
+        // Get a value content type header if any.
+        let header = response.headers().get(header::CONTENT_TYPE);
+        let header = header.and_then(|value| value.to_str().ok());
+
+        match header {
+            Some("text/markdown") => Some(ContentType::Markdown),
+            Some("image/png") | Some("image/jpeg") => Some(ContentType::Image),
+            Some("application/pdf") => Some(ContentType::Pdf),
+            _ => None,
```
oberien: Could we try and extract the content-type here similar to plain files from the extension if no `ContentType` was provided?

src/resolve/source.rs
```diff
@@ -101,7 +102,23 @@ impl Source {
                 let parent = path.parent().unwrap().to_owned();
                 to_include(path, Context::LocalAbsolute(parent))
             }
-            SourceGroup::Remote => unimplemented!(),
+            SourceGroup::Remote => {
+                let downloaded = match remote.http(url) {
+                    Ok(downloaded) => downloaded,
+                    Err(RemoteError::Io(io)) => return Err(io),
+                    Err(RemoteError::Request(req)) => return Err(io::ErrorKind::ConnectionAborted.into()),
```
oberien:
```suggestion
                    // TODO: proper error handling with failure
                    Err(RemoteError::Request(req)) => return Err(io::ErrorKind::ConnectionAborted.into()),
```

oberien (changes requested):
With the `ContentType` enum and my comment about falling back to deriving the content-type from the file ending, the content-type detection via file ending should be a separate function used by both `remote.rs` as fallback and `source.rs` for non-remote sources.

src/resolve/remote.rs
197g: By this method, we first unlink the original file and create a new inode. Then someone who still holds on to the older file handle can read from it undisturbed.

src/resolve/remote.rs
197g: You're right, this transformation needs to be injective. `file_stem` is None when the path contains no file name, so maybe this should be an `expect` or `map_err()?` instead.

src/resolve/remote.rs
oberien: Good point. Probably worth a comment.

src/resolve/remote.rs
oberien: Currently I can't think of a use-case where you'd include a file without a name and extension. So an expect is better here for now until someone really needs it.

src/resolve/remote.rs
197g: This is already the case, see https://github.com/oberien/pundoc/pull/4/files/64b7277239f2bfbbfb0d1d2b3fd1b5bff5bbf21a#diff-827d0799743e13fc0eb5bb26f73f89e8R119

I'll add a comment explicitly specifying that behaviour, and document `to_include` accordingly.

src/resolve/mod.rs
oberien: Definitely clearer now.

src/resolve/remote.rs
```diff
@@ -0,0 +1,148 @@
...
+        //
+        // Also, '+' is a reserved character that can not appear unescaped.
+        let path = url.path().replace('/', "+");
+        let path = Path::new(&path);
+
+        // Since pdflatex is picky with file extensions, replace all dots.
+        let stem = path.file_stem()
+            // `path` already was valid utf-8
+            .map(|osstr| osstr.to_str().unwrap())
+            // file_stem is a part of the file_name, which exists if the last component is not `..`
```
oberien: This case can also happen if you just provide a link like `http://foo.bar/baz/` (which internally resolves to `index.html`, but that isn't part of the URL).

src/resolve/mod.rs
197g: It was a copy from `assert_eq!` before, so completely bloated.

src/resolve/remote.rs
197g: Redirects are currently off completely for exactly this reason. In case of a directory, the `file_stem` is the directory name if I read the doc correctly. I'll add tests for there since I'm not sure how the top-level reacts.

oberien:
@HeroicKatora Are you going to perform the last few changes or should I implement them?

197g:
On Friday at the earliest, I was a bit short on time. But I'll do it then if it is still open.

src/resolve/remote.rs
197g: Found two pairs of different urls that map to the same file paths. Tests have been added and will be fulfilled tomorrow.

src/resolve/remote.rs
197g: In some first impression, the `mime` crate complicates this match expression. But it could help by also allowing media types with added parameters such as text encoding.

src/resolve/remote.rs
197g: Apparently, `set_extension("")` does nothing, inconsistent with `extension` returning `Some("")` when the filename ends in a `.`

197g: Okay, should be fixed now by simply performing the extension with extension manually. Since the string was utf-8 encoded anyways, there should be no issue.

src/resolve/remote.rs
```diff
@@ -80,45 +82,53 @@ impl Remote {
...
-            Some("image/png") | Some("image/jpeg") => Some(ContentType::Image),
-            Some("application/pdf") => Some(ContentType::Pdf),
+        let header = response.headers().get(header::CONTENT_TYPE)
+            .and_then(|raw| raw.to_str().ok())
+            .and_then(|string| string.parse::<mime::Mime>().ok())?;
+
+        match (header.type_(), header.subtype()) {
+            (mime::TEXT, sub) if sub == "markdown" => Some(ContentType::Markdown),
+            (mime::IMAGE, mime::PNG) | (mime::IMAGE, mime::JPEG) => Some(ContentType::Image),
+            (mime::APPLICATION, sub) if sub == "pdf" => Some(ContentType::Pdf),
```
oberien: Couldn't you just check against `mime::APPLICATION_PDF` and `mime::IMAGE_*` instead of manually comparing against strings?

src/resolve/remote.rs
oberien: Good catch and fix

src/resolve/remote.rs
```diff
@@ -80,45 +82,53 @@ impl Remote {
...
 
         // http(s) domains must contain a hostname
         target.push(url.host_str().unwrap());
 
-        // http(s) must not be cannot-be-base
-        //
-        // Also, '+' is a reserved character that can not appear unescaped.
-        let path = url.path().replace('/', "+");
+        // http(s) must not be cannot-be-base.  Also, '+' is a reserved character that can not
+        // appear unescaped and "+0" is a unique sequence to this replacement.
```
oberien: I'm not too sure about this guarantee. The URL spec doesn't have this behaviour, any `+` should be treated literally. Only the HTML spec says that the URL should be application/x-www-form-urlencoded, which ensures the behaviour you're stating. See [this stackoverflow post](https://stackoverflow.com/questions/1005676/urls-and-plus-signs).

oberien (changes requested):

src/resolve/remote.rs
197g: Good point, `+` is only a reserved character to be used as a sub-delimiter. I misread the spec.

oberien: Why don't you just SHA3 the URL and keep the file ending if it exists? Not only will that get rid of all parent-child structures so you don't need to manually handle them, but also generate a unique hash for every possible URL (not considering collisions obviously).

src/resolve/remote.rs
197g: Coverage of standardized mime types seems to be spotty at best, `text/markdown` should be standard too but I did not find it in the list of constants. The lib does not actually solve comparison well, it's not simply possible to assert that the type matches any `image/*` afaik. See https://github.com/hyperium/mime/issues/73 and https://github.com/hyperium/mime/issues/52#issuecomment-324426832

src/resolve/remote.rs
197g: Hm, it might be the best course of action. I was reluctant to throw away the path information in case we want to manually construct the path from another url (e.g. redirects or similar) but it's probably better to define a set of set of 'unique' parameters anyways. The same url could refer to different resources depending on request+response headers. Yeah, let's encapsulate this unique identifying information properly (and use hashes).

src/resolve/remote.rs
```diff
@@ -80,56 +93,64 @@ impl Remote {
...
+            .and_then(|string| string.parse::<Mime>().ok());
 
-        match (header.type_(), header.subtype()) {
-            (mime::TEXT, sub) if sub == "markdown" => Some(ContentType::Markdown),
-            (mime::IMAGE, mime::PNG) | (mime::IMAGE, mime::JPEG) => Some(ContentType::Image),
-            (mime::APPLICATION, sub) if sub == "pdf" => Some(ContentType::Pdf),
-            // Let the file extension logic take over.
-            _ => None,
-        }
+        Self::from_parts(response.url().clone(), mime)
```
oberien: s/.clone()/.to_owned()/
That makes it clearer that you have a reference here from which you want the owned version. With `.clone()` I was just wondering why you'd clone something in the very last line of a function.

src/resolve/remote.rs
```diff
@@ -80,56 +93,64 @@ impl Remote {
...
+        let file_hash = Sha256::new()
+            .chain(key.path.to_str().unwrap())
+            .chain(key.mime.as_ref().map(Mime::as_ref).unwrap_or(""))
+            .result();
+
+        let mut hash_name = String::new();
+        file_hash.into_iter().for_each(|x| {
+            hash_name.push(std::char::from_digit((x >> 4).into(), 16).unwrap());
+            hash_name.push(std::char::from_digit((x & 15).into(), 16).unwrap());
+        });
```
oberien: `GenericArray<u8, _>` implements `LowerHex` and `UpperHex`, so we can just use that here:

```rust
let mut hash_name = format!("{:x}", file_hash);
```

src/resolve/remote.rs
```diff
@@ -80,56 +93,64 @@ impl Remote {
...
+
+        if let Some(extension) = extension {
+            hash_name.push('.');
+            hash_name.push_str(extension.to_str().unwrap());
+        }
+
+        let mut target = self.temp.clone();
+        target.push(&key.host);
```
oberien, line 114: Why don't you just include the target in the hash? Why do you create a new folder for each domain? I guess that could make debugging better as you don't have to open every single file to find out the one which seems buggy.

oberien (approved):

src/resolve/remote.rs
197g: Pretty much for that reason. I personally wouldn't like it to be a complete blackbox. It's debatable but a bit of a bikeshedding issue.

src/resolve/remote.rs
197g: Instead we can make `from_parts` take `&Url`. All relevant components are cloned individually.

PR #15, opened 2018-12-05
[WIP] Tracking Beamer

197g:
- [ ] Remove the stupid hack for creating beamer frames

All headers nested 3 levels or deeper are treated as frames for the sake of beamer. Use of `subsubsection` is discouraged anyways. We could switch to `subsubsubsection` as well, that is not built into the document class afterall, but it makes creating frames less pretty. Beamer also has an internal labeling system as frame attributes that should probably be used instead of the raw `\label` mechanism.

The implemention abuses a `Tag::Header` context to keep tack of beamer frames and close their environment properly. This is an absolutely abhorrent hack, and it means there needs to be an additional non-frame header at the end of the document. 

Something like this should instead be built into the design for document classes (`Backend` impls). I'm very sorry for this being included at all but I need it kind of right now, and I didn't have time to look into the overall structure of the backend for doing the necessary rework.

- [ ] Support beamer options

Also styles and colours, all deeper customizations can be done in the document.

- [ ] Allow more control over frame options

Some way to control whether `fragile` is set, or which mechanism to use for overlays. Also, `plain` is likely useful. We could also decide to forward to pure latex code in the markdow file here.

### DIFF
--- a/src/backend/latex/document/beamer.rs
+++ b/src/backend/latex/document/beamer.rs
@@ -1,0 +1,110 @@
+use std::io::{Write, Result};
+use std::fs;
+use std::path::Path;
+
+use typed_arena::Arena;
+
+use crate::backend::Backend;
+use crate::backend::latex::{self, preamble};
+use crate::config::Config;
+use crate::generator::Generator;
+
+#[derive(Debug)]
+pub struct Beamer;
+
+impl<'a> Backend<'a> for Beamer {
+    type Text = latex::TextGen;
+    type FootnoteReference = latex::FootnoteReferenceGen;
+    type Link = latex::LinkGen;
+    type Image = latex::ImageGen;
+    type Pdf = latex::PdfGen;
+    type SoftBreak = latex::SoftBreakGen;
+    type HardBreak = latex::HardBreakGen;
+    type TableOfContents = latex::TableOfContentsGen;
+    type Bibliography = latex::BibliographyGen;
+    type ListOfTables = latex::ListOfTablesGen;
+    type ListOfFigures = latex::ListOfFiguresGen;
+    type ListOfListings = latex::ListOfListingsGen;
+    type Appendix = latex::AppendixGen;
+
+    type Paragraph = latex::ParagraphGen;
+    type Rule = latex::RuleGen;
+    type Header = latex::HeaderGen;
+    type BlockQuote = latex::BlockQuoteGen;
+    type CodeBlock = latex::CodeBlockGen;
+    type List = latex::ListGen;
+    type Enumerate = latex::EnumerateGen;
+    type Item = latex::ItemGen;
+    type FootnoteDefinition = latex::FootnoteDefinitionGen;
+    type Table = latex::TableGen;
+    type TableHead = latex::TableHeadGen;
+    type TableRow = latex::TableRowGen;
+    type TableCell = latex::TableCellGen;
+    type InlineEmphasis = latex::InlineEmphasisGen;
+    type InlineStrong = latex::InlineStrongGen;
+    type InlineCode = latex::InlineCodeGen;
+    type InlineMath = latex::InlineMathGen;
+    type Equation = latex::EquationGen;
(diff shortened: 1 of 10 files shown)


oberien:
> All headers nested 3 levels or deeper are treated as frames for the sake of beamer. Use of `subsubsection` is discouraged anyways. We could switch to `subsubsubsection` as well, that is not built into the document class afterall, but it makes creating frames less pretty.

I'd like to treat this like pandoc i.e. my regex replace script for pandoc handles headers → frames.
I'd like to let each slide have a title and a subtitle with the title being the section and the subtitle being the subsection. Thus, each subsection should result in a new page. Each subsubsection should create a beamerboxrounded.
The general frame template I use is therefore this:

```latex
\section{Section}
\subsection{SubSection}
\begin{frame}
\frametitle{\insertsection}
\framesubtitle{\insertsubsection}
...
\end{frame}
```

> Beamer also has an internal labeling system as frame attributes that should probably be used instead of the raw `\label` mechanism.

Given the above frame template I've had success with just annotating the `\section` command with the usual `\label`.

> The implemention abuses a `Tag::Header` context to keep tack of beamer frames and close their environment properly. This is an absolutely abhorrent hack, and it means there needs to be an additional non-frame header at the end of the document.

I'm amazed and disgusted by this idea at the same time.

> Something like this should instead be built into the design for document classes (`Backend` impls). I'm very sorry for this being included at all but I need it kind of right now, and I didn't have time to look into the overall structure of the backend for doing the necessary rework.

I'm not sure how to handle this properly. I've thought about it for the last day and my best idea is to make the struct implementing `Backend` be able to hold state. Then in addition to the `CodeGenUnit` we have a `StateCodeGenUnit` (name open for discussion), which gets the `Backend` as first argument in its `new` method.  
Every type in the `Backend` trait currently having the trait bound `CodeGenUnit` would then be changed to taking a `StateCodeGenUnit` instead. To keep backwards compatibility, we have a generic `impl<T: CodeGenUnit> StateCodeGenUnit for T` just as I did with `SimpleCodeGenUnit` and `MediumCodeGenUnit` (names also open for discussion).

That way the `Header` implementation could save its state in the `Backend` implementation.

> Some way to control whether `fragile` is set, or which mechanism to use for overlays. Also, `plain` is likely useful. We could also decide to forward to pure latex code in the markdow file here.

While I do agree that inline-latex could be used here, one goal is to also have HTML output at some point in the future. At that point inline-latex would be rendered incorrectly, which is why in the long term I'd prefer to abstract away all inline latex with commands / custom syntax. For the time being this is probably out of scope.

Additionally I'd like to have the following features for beamer:

* [ ] Hline (`---`) should create a new frame with the same header
* [ ] Easy support for multiple columns (don't know which syntax to use here)
 ```latex
\begin{columns}
\begin{column}{.6\textwidth}

\end{column}
\begin{column}{.38\textwidth}

\end{column}
\end{columns}
```
* [ ] tightlist for all lists (maybe configurable, probably not as it's a sane default)

197g:
> I'd like to let each slide have a title and a subtitle with the title being the section and the subtitle being the subsection. Thus, each subsection should result in a new page.

Just a note on this. The general structure is not my own idea but follows the guidelines mentioned in the [`beamer` user guide section 10.2](http://tug.ctan.org/macros/latex/contrib/beamer/doc/beameruserguide.pdf). In particular, sections and subsections create not only entries for the table of contents but also introduce navigation anchors. Thus I'm slightly wary of using depth 1 and 2 as frame delimiters as well.

Agree on most other points though, multi column etc would be an awesome included feature also for poster presentations.

oberien:
I was following the way [pandoc handles slides](https://pandoc.org/MANUAL.html#producing-slide-shows-with-pandoc) which I experienced to be intuitive and easy to use. Pandoc defines a configurable `slide-level` (default 2). Headings above that level result in a title-slide, headings at that level create a slide, and headings below that level create sub-headings on the slide (which with defaults at level 3 is a `\beamerboxrounded`).

oberien:
Closing in favour of #97.

Issue #20, opened 2019-02-11
Move Url → Include parsing to broken_link_callback


oberien:
Currently the broken link callback accepts all links. Shortcut links can be:
1. a proper CommonMark shortcut link `[https://example.com]`
2. a biber reference `[@foo]`
3. a reference (InterLink) `[#interlink]`
4. something for the resolver `[include foo.md]`, `[TOC]`, …
5. something invalid, like `[nonexistent]`

The frontend parses (2) and (3), but passes through all other shortcut links, because the generator handles and translates case (4). This structure results in invalid / nonexistent links to be passed through to the backend, which accepts those (5) links as (1) instead of as text: `[nonexistent]` → `\url{nonexistent}`.

As these conversions are context free, they should be performed in the frontend. Links (1) aren't passed through to the broken_link_callback in the first place. Cases (2), (3) and (4) should be accepted by the broken_link_callback, while case (5) should be rejected by the broken_link_callback.
Then the frontend only needs to handle cases (1)-(4) and not (5), which it does currently. The frontend can then convert (4) to `Include` instead of to `Url`, which moves those conversions from the generator to the frontend.

oberien:
With this change the `refsolve` (`broken_link_callback`) should be moved to the frontend just as the `pulldown_cmark::Parser` creation and initialization.

Issue #18, opened 2019-02-10
What to do with Link / Image Title and Alt-Text?


oberien:
With #16 there is a common and defined way for all elements to have config, including a label, and captions for figures. The old idea of an image's title was to be used as label and its alt-text as config. This may now be obsolete, especially because the config can contain a label.

So what shall we do with the title of links and images, and the alt-text of images?

moritzuehling:
My personal recommendation would be to actually use those features in the way they are intended:

- Use the alt text as an accessibility feature (for example, as documented [in this SO answer](https://tex.stackexchange.com/questions/75102/need-alt-text-for-images-in-pdf-for-screen-readers/75104#75104).
- Use the title as a tooltip using [pdftooltip](https://tex.stackexchange.com/a/75151)

This way the features are very compatible with normal markdown, and you give people a way to be nice.

Issue #11, opened 2018-10-24
Project Name


oberien:
**Decision: heradoc**

TODO:

- [ ] `s/pundoc/heradoc/g` ∀ files
- [ ] rename github repository
- [x] update github description (if needed)
- [ ] update repository urls in local git copies

This project has reached a state where we need to think about a possibly different name. If we ever want this tool to be used by more than 2 people, the ambiguity between `pundoc` and `pandoc` is too high. Here are some suggestions with their explanations:

1. `pundoc`: "pun" wrt. "pandoc" is pretty obvious. Unfortunately, they are very ambiguous, which is why we should look for something different.
2. `omnidoc`: "pan-" is a combining form for "all-", taken from the Greek. "omni-" is also a combining form with the same meaning, but it stems from Latin.
There is already a product called `omnidoc`, which is some medical device for documentation of gel. "OMNIDOC" and "OmniDoc" are registered trademarks in the US and KR respectively. "omnidoc" thus shouldn't be a problem, albeit maybe harder to search for.
3. `heradoc`: Pan is the Greek god of shepherds. Hera is a Greek goddess, the queen of gods. Hera > Pan, which means heradoc > pandoc ;) Furthermore, heradoc sounds a little like heretic, which would fit with the general view of the rust community regarding RIIR, which is exactly what we're doing here.
Heradoc is not a registered trademark, the closest to it would be THERADOC, so we're fine on that front. Heradoc *is* a German app for agriculture documentation though. However, I don't think that'll pose too much of an issue.
4. `maydoc`: "PAN-PAN is the international standard urgency signal that someone […] uses to declare that they have a situation that is urgent but, for the time being at least, does not pose an immediate danger to anyone's life or to the vessel itself." (thanks wikipedia) Similar is the word "Mayday", which is a stronger urgency signal requesting immediate highest priority. Thus `PANdoc` vs `MAYdoc`. Advantages are internationally equal pronunciation, two syllables (compared to 2 and 3 having three) and it's a soft word that's easier to say.
Maydoc is not trademarked in any way, there isn't even a similar existing trademark. On Google the first hit is a youtube channel with 0 subscribers, so the word is completely free for our use.

Here is a table, collecting all votes so far.

omnidoc | heradoc | maydoc
--- | --- | ---
3 | 2 | 3

197g:
[...]

PR #22, opened 2019-02-12
Labels

### LINKED ISSUE #16: Labels, References, Element-Config and Figures
* currently label and reference systems are inconsistent with each other
* somewhat enforced label prefix
    * actually not really
    * complicates system unnecessarily
* labels are generated in backend
    * no syntax enforced
    * could result in different label syntax in different backends

TODOs:

- [x] References rework (frontend)
- [x] cskvp parsing (single values and key=value pairs)
- [x] Refactor backend to destructure its argument instead of accessing all of its values.
  This makes adding new members to a struct result in an error everywhere, which requires us integrating it everywhere.
- [x] Labels in all elements (frontend)
    - [x] Label parsing
    - [x] Remove label parsing from backend
- [x] Figures
    - [x] `figures` config option
- [x] Element-Config in all elements

## New system suggestion:

Label: Definition of label
Reference: Reference to a label

References:

* 2 Namespaces:
    * Biber (starts with `@`)
        * biber single: `[@foo]`, `[@foo 42-69, 322]`, `[@foo Chapter 13]`
            * `BiberSingle(Cow, Option<Cow>)`
        * biber multiple: `[@foo 21-23,25, @bar Chapters 10-15]`
            * `BiberMultiple(Vec<(reference, Option<attributes>)>)`
    * Links (starts with `#` or `^`)
        * all links are equal
        * Footnotes: `[^foo]`
            * part of Links with prefix `fnote:`
            * `^foo` is translated to `fnote:foo`
            * should still be referenced as `[^foo]`, otherwise it's a cref instead of a footnotemark
        * cref: `[#foo]`
            * `#foo` is just `#foo`
            * `InterLink(LabelReference)` (uppercase=false)
        * Cref: `[#Foo]`
            * `InterLink(LabelReference)` (uppercase=true)
        * cref / Cref with content: `[text][#foo]`, `[text](#foo)`
            * `InterLinkWithContent(LabelReference)`
        * Text: `[text][doesn't start with # or ^]`
            * output as plain text
            * issue warning (blocked on #12)
        * anything else: `[foo.md]`, `[anything not captured]`
            * `Url`
        * anything else with content: `[text](file.md)`, `[text](anything not captured)`
            * `UrlWithContent`

Labels:

* generated and added to elements in frontend
    * common label syntax
* Fotnotes: `[^foo]: Text for footnote`
    * translated to `fnote:foo`
* Automatic labels for each section:
    * `# Heading Foo$() Bar` translated to `heading-foo-bar`
    * all chars converted to lowercase
    * `" "` converted to `"-"`
    * `[^a-zA-Z0-9-_ ]` converted to `""`
* Inline labels: `{#foo}`
    * translated to label `foo`
* Labels after section: `# Heading {#label-for-heading}`
    * translated to `label-for-heading`
* Prefix-style label must be at the beginning of a paragraph (empty line before) and either end the paragraph (empty line after) or be followed by a SoftBreak (line break)
* Labels before Image / Table / Code / Graphviz (empty line required before table):
  ```md
  {#label,config...}

  Heading A | Heading B
  --- | :-:
  Cell 1 | Cell 2
  ```
* Labels for Code / Graphviz, different way (for easier compatibility with other MD parsers):
  ````md
  ```language,#foo,config...
  fn main() {}
  ```
  ````
* Labels before Image (can't be on its own paragraph, so must be followed by linebreak followed directly by image):
  ```md
  {#fancy-image}
  ![alt-text](link "title")
  ```

Element-Config:

* Config: cskvp (comma-seperated key-value pairs)
    * all: 
        * label: `#foo`
        * `figure=true` / `figure=false`
            * Figure Caption: `caption=`
    * Image / Graphviz: `scale=`, `width=`, `height=`
    * Code
    * Table

Figure handling:

* Add new config option `figures=`
    * `true`: images / graphviz / code / tables default to being wrapped in figures
    * `false`: images / graphviz / code / tables default to being inline
* if `figure=true` is specified (or `figure=false` is not specified but `figures=true` configured in the config):
    * put label within figure (instead of within element)
    * wrap element in Figure Tag
    ```rust
    Event::Start(Tag::Figure(label=…, …))
    // item
    Event::End(Tag::Figure(…))
    ```

Elements with Config:

* Have different possible config within each struct

```rust
struct Image {
    …,
    label = Cow,
    scale = …,
    width = …,
    height = …,
}
```

oberien:
Fix #16 

TODO:

- [x] Merge master into labels (scrreprt)


src/backend/latex/mod.rs
```diff
@@ -35,4 +40,47 @@ use self::complex::{
     GraphvizGen,
 };
 
+fn inline_figure_begin(out: impl Write, label: &Option<Cow<'_, str>>, caption: &Option<Cow<'_, str>>) -> Result<()> {
```
197g: These could benefit greatly from being written as one struct `Figure<'a>` with associated methods to write being and end. It would also make it easier to add things as specifying the placement later on. Proposing public members but useful constructors and ownership over the `Cow<'a, _>`, that should be enough to encapsulate properly while maintaining compat. where the label is needed after use.

src/cskvp.rs
```diff
@@ -0,0 +1,131 @@
...
+        let figure_double = double.remove("figure")
+            .and_then(|s| match s.parse() {
+                Ok(val) => Some(val),
+                Err(_) => {
+                    // TODO: warn
+                    println!("cannot parse `figure={}`, only `true` and `false` are allowed", s);
+                    None
+                }
+            });
+        let figure = single.remove_element(&"figure").is_some();
```
197g: I'd rather just ignore additional figure declarations, and take the last one consistent with `label`:

```
let figure = Some(single.remove_element(&"figure").is_some());
let nofigure = Some(single.remove_element(&"figure").is_some());

let mut figure = [figure_double, figure, nofigure].into_iter().flatten();
let figure = figure.next();
let figure = figure.fold(figure, |_, replace| {
    // warn about multiple.
    replace
});

File: src/frontend/mod.rs [hunk 3]
```diff
@@ -62,212 +49,417 @@ impl<'a, B: Backend<'a>> Frontend<'a, B> {
...
+        // check if code is math mode
+        let mut text = match self.parser.next().unwrap() {
             CmarkEvent::Text(text) => text,
-            e => unreachable!("InlineCode should always be followed by Text but was fallowed by {:?}", e),
+            CmarkEvent::End(CmarkTag::Code) => {
+                self.buffer.push_back(Event::Start(Tag::InlineCode));
+                self.buffer.push_back(Event::End(Tag::InlineCode));
+                return;
+            }
+            e => unreachable!("InlineCode should always be followed by Text or End(Code) but was fallowed by {:?}", e),
```
197g: Great chance to fix a typo: `followed`!

src/frontend/mod.rs
```diff
@@ -62,212 +49,417 @@ impl<'a, B: Backend<'a>> Frontend<'a, B> {
...
-        let res = match single[0] {
-            "equation" | "math" | "$$" => {
-                self.state = State::Equation;
-                Event::Start(Tag::Equation)
+
+        // check if language has label/config
+        let language;
+        if let Some(pos) = lang.find(',') {
+            language = &lang[..pos];
+            if cskvp.is_some() {
```
197g: Unclear that this will always empty the cskvp:

```
if let Some(cskvp) = cskvp.take() {
}
```

src/frontend/mod.rs
```diff
@@ -62,212 +49,417 @@ impl<'a, B: Backend<'a>> Frontend<'a, B> {
...
+                // TODO: error
+                println!("Code has both prefix and inline style labels / config, ignoring both");
+                // don't print warnings about unused properties
+                // will be cleaned up as it's on the stack anyways
+                mem::forget(cskvp.take());
```
197g: Maybe `cskvp` should have a `drain(&mut self, impl Write)` instead, then drain to intermediate streams or `io::sink`. And warn about internal bug potentially in the destructor? Not completely sure, but seems less prone to warning bugs.

src/frontend/mod.rs
```diff
@@ -62,212 +49,417 @@ impl<'a, B: Backend<'a>> Frontend<'a, B> {
...
+                let cskvp = Cskvp::new(&lang[pos+1..]);
+                // check for figure and handle it
+                self.handle_cskvp(cskvp, CmarkEvent::Start(CmarkTag::CodeBlock(Cow::Borrowed(language))));
+                return;
+            }
+        } else {
+            language = lang;
+        }
+
+        let tag = match language {
```
197g, line 229: Rather than all the `and_then` stuff, maybe assume an empty `cskvp` instead. Would necessitate `Cskvp: Default` or something to that effect.

```
let cskvp = cskvp.unwrap_or_else(Cskvp::default);
```

src/frontend/mod.rs
```diff
@@ -62,212 +49,417 @@ impl<'a, B: Backend<'a>> Frontend<'a, B> {
...
+        // If neither the first or the second are specified, we use the default one.
+        // Otherwise we take the one that's specified.
+        let prefix = cskvp.as_mut().and_then(|cskvp| cskvp.take_label());
+        // Consume elements until end of heading to get its text.
+        // Convert them and put them into the buffer because the're still needed.
+        let current_index = self.buffer.len();
+        let text = self.convert_until_end_inclusive(|t| if let CmarkTag::Header(_) = t { true } else { false });
+
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r"\{(#[a-zA-Z0-9-_]+\})\w*$").unwrap();
```
197g: The closing curly is part of the capture group while the opening is not? Seems weird, a comment what the regex want to extract would be good anyhow.

197g (changes requested):

src/frontend/mod.rs
oberien: I don't fully understand what you mean here. What would `drain` write into? My idea was that I *don't* need to call a function to print the warning about possibly unused attributes everywhere. I do agree that `mem::forget` here seems weird, so maybe I should add a method which consumes everything and thus not printing a message. But then again `mem::forget` does pretty much the same.

src/frontend/mod.rs
oberien: Thank you! First I found my `.as_mut…` solution fancy, but copy-pasting it a thousand times felt really bad.

src/frontend/mod.rs
197g: > so maybe I should add a method which consumes everything and thus not printing a message.

I meant pretty much the same, I think:

```
impl<'a> Cskvp<'a> {
    /// Debug and then remove all remaining, unused attributes into the writer.
    pub fn drain(&mut self, o: impl Write) -> io::Result<()> {
        if let Some(label) = self.label.take() {
            writeln!(o, "label ignored: {}", label)?;
        }
        if let Some(figure) = self.figure.take() {
            writeln!(o, "figure config ignored: {}", figure)?;
        }
        if let Some(caption) = self.caption.take() {
            writeln!(o, "caption ignored: {}", caption)?;
        }
        for (k, v) in self.double.drain() { // here we already have drain!
            writeln!(o, "Unknown attribute `{}={}`", k, v)?;
        }
        for attr in self.single.drain(..) {
            writeln!(o, "Unknown attribute `{}`", attr)?;
        }
        Ok(())
    }
}

impl<'a> Drop for Cskvp<'a> {
    fn drop(&mut self) {
        self.drain(stdin().lock()).unwrap();
    }
}
```

with usage:

```
// replaces mem::forget(), to just clear any unused parts.
cskvp.drain(io::sink()).unwrap()
```

Improvement: `mem::forget` will currently leak the memory of the hashmap and the vec, and could potentially leak `Cow::Owned` when parsing gets improved to handle quote escapes.

oberien: > `mem::forget` will currently leak the memory of the hashmap and the vec, and could potentially leak `Cow::Owned` when parsing gets improved to handle quote escapes.

Ouch, totally forgot about that. That's no good. Definitely requires a change!

oberien: Unfortunately that solution doesn't really work once we use the `log` crate, because we'd like to have the `warn!` output's prefix in each printed line as well. So the `drain` method should just `take()` all options and `clear()` the Vec and HashMap, while the `drop` method stays as it is. The `drain` method can still consume `self`, making that behaviour clearer.

197g: Ah yes. Alternatively, could specify a new trait for the `drain` operation with 5 methods for the different unhandled items, implemented for `T: io::Write` but in the future also for a private `struct Warn` that gets called upon in `Drop`. That may be convoluted though.

oberien: I ended up writing a `.clear()` method and am calling that to reset the `cskvp` to its `Default` state.

oberien:
What is your verdict to the fixes to the other comments and the PR in general?

197g:
Positive:
* `rust,#label` is pretty intuitive to label code blocks and figures
* parity with links to refs is pretty cool
* *optional* `figure` around everything is nice
* unified parsing is definitely necessary and much cleaner to implement now.

Ehhh:
* Forced paragraph before tables feels unatural but I can live with it until a better syntax is there
* Just don't publish the tool widely before ;)
* The exact parts in `figure` can be figured out in a later implementation, after use. It's good enough and I don't see an obvious weakness in usability from markdown perspective.

Negative:
* Tightens the dependency of backend and frontend through the figure nesting etc. But that should not be a problem for most possible targets I can imagine.

src/backend/latex/mod.rs
197g: Make this a note in improvements for TODO somewhere, for code clarity and to enhance/clarify the feature set of figure. Then I'm content with the situation right now.

oberien: I did implement your suggestion in a slightly different way: https://github.com/oberien/pundoc/pull/22/files#diff-3eae75b8fbad57dfe22fc3bbf684fa36R43

197g: I see, I'd have used an enum with `fn env(&self) -> &'static str` instead of stringly typed `figure`/`table`, in case we want to add more? I didn't consciously use the `float` package yet ^^

197g (approved):

oberien:
> unified parsing is definitely necessary and much cleaner to implement now.

What do you mean with "unified parsing"?

> Forced paragraph before tables feels unatural but I can live with it until a better syntax is there

Tables aren't a part of CommonMark and the way pulldown-cmark parses it requires any table to be preceded by an empty line. As such we can't really do anything against that.

> Just don't publish the tool widely before ;)

Actually my plan was to announce this tool on reddit once #16, #18, #20, #23 and #11 are done. As I said I don't think we really have a choice regarding tables and their figures / labels.

> The exact parts in `figure` can be figured out in a later implementation, after use. It's good enough and I don't see an obvious weakness in usability from markdown perspective.

> Tightens the dependency of backend and frontend through the figure nesting etc. But that should not be a problem for most possible targets I can imagine.

My idea is the following: People should be able to use the floating environment, which they can specify with `figure/nofigure/figure=true/false`. In that case the optional `label` and `caption` are applied to that figure and the figure converted into its own Start/End Tag. Unfortunately you shouldn't use `\begin{figure}` for tabulars, but use `\begin{table}` instead. As such there is `Tag::Figure` and `Tag::TableFigure`. Maybe we could rename those to `Floating` and `TableFloating` to be more general instead of using latex-lingo.
People should also be able to specify a label / caption without a floating environment. In that case the label / caption is applied to the element directly as struct fields. In latex some elements can use those directly (like `\begin{lstlisting}[label=foo,caption=bar,language=rust]`), while others just don't work. If a label is specified before / after an image, it references the section instead of the image. As such I created the hack of creating a figure with position `H`, which the caption and label will be applied to if specified.

src/backend/latex/mod.rs
oberien: I encapsulated it by making that string private and only having it in the creation method name. It's more straightforward and provides pretty much the same guarantees.

197g:
> What do you mean with "unified parsing"?

String escapes working the same everywhere, or with a non-handrolled format maybe. Cleaner separation now that it is its own module.

> People should also be able to specify a label / caption without a floating environment. In that case the label / caption is applied to the element directly as struct fields. In latex some elements can use those directly (like `\begin{lstlisting}[label=foo,caption=bar,language=rust]`), while others just don't work. If a label is specified before / after an image, it references the section instead of the image. As such I created the hack of creating a figure with position `H`, which the caption and label will be applied to if specified.

It just seemed unclear to me if pundoc should actively prevent you from trying to specify invalid labels, or labels at confusing positions. Regarding the elements that inline captions also creates issues where a `]` *or* a `}` within a lable may be disallowed or create absolutely confusing errors in pdflatex. The same would not be true for html backend for example. These sort of differences may be important and it is not clear to me how we should resolve this to be both future compatible and open enough for hacking where necessary (e.g. we should just escape any use of brackets as commands may be useful).

Edit: That said, this PR doesn't worsen the situation, it only makes it more apparent because labels are much more comfortable to use in other regards.

oberien:
> String escapes working the same everywhere, or with a non-handrolled format maybe. Cleaner separation now that it is its own module.

Good point, I created #24.

> It just seemed unclear to me if pundoc should actively prevent you from trying to specify invalid labels, …

My plan is to try to have a consistent frontend, which is compatible with all backends, such that any one markdown file can be rendered with every backend easily. As such, for example labels should match `#([a-z\-_]+)`.

This is a lot of bikeshedding though and probably requires quite a huge PR to find a code structure which allows easy verification of all attributes, element, and anything else that we might not think about right new. Therefore my current consensus is to keep the 0.1.0 version minimal in that regard and allow the user full control / responsibilities. Once we have some experience with users and other use-cases, this will become easier to judge.

> That said, this PR doesn't worsen the situation

:+1:

PR #26, opened 2019-02-25
Move Shortcut link command parsing from generator to frontend

### LINKED ISSUE #20: Move Url → Include parsing to broken_link_callback

oberien:
Fix #20

### DIFF
--- a/src/backend/latex/simple.rs
+++ b/src/backend/latex/simple.rs
@@ -30,7 +30,11 @@ impl<'a> MediumCodeGenUnit<Cow<'a, str>> for TextGen {
                 '_' if in_inline_code => s.push_str("\\char`_"),
                 '_' if !in_code_or_math => s.push_str("\\_"),
                 '#' if in_inline_code => s.push_str("\\#"),
+                '#' if !in_code_or_math => s.push_str("\\#"),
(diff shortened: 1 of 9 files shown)


197g:
Reviewing Thursday, the other one as well. Need some time for uni stuff.

Issue #19, opened 2019-02-11
cskvp: Support double-quoted keys and values


oberien:
Currently keys and values can't contain commas or spaces. This can be fixed by allowing double-quoted strings as keys and values like `"My complicated\" key"="Knock, knock"`.

oberien:
@HeroicKatora This is actually not as simple as expected. In Commonmark Text

> Any ASCII punctuation character may be backslash-escaped
> -- <https://spec.commonmark.org/0.28/#backslash-escapes>

> An ASCII punctuation character is !, ", #, $, %, &, ', (, ), *, +, ,, -, ., /, :, ;, <, =, >, ?, @, [, \, ], ^, _, `, {, |, }, or ~.
> -- https://spec.commonmark.org/0.28/#ascii-punctuation-character

As such, a label / caption cskvp can't just single-backslash-escape quotation marks. For example the markdown text `{#label, caption = "caption with \" quotes"}` will be translated to (and passed to us as) `{#label, caption = "caption with " quotes"}`. The only way I think we can "fix" this is by requiring double-quoted strings to escape inner quotes with two backslashes: `{#label, caption = "caption with \\" quotes"}`.

PR #28, opened 2019-03-02
Quoted strings

oberien:
Unresolved question in <https://github.com/oberien/pundoc/issues/19#issuecomment-468943142>.


src/cskvp.rs
```diff
@@ -138,3 +137,129 @@ impl<'a> Drop for Cskvp<'a> {
...
+    }
+
+    fn next_unquoted(&mut self, delimiters: &[char]) -> Cow<'a, str> {
+        let idx = delimiters.iter().cloned()
+            .filter_map(|delim| self.rest.find(delim))
+            .min()
+            .unwrap_or(self.rest.len());
+        let mut rest = self.rest.split_off(idx);
+        mem::swap(&mut rest, &mut self.rest);
+        let mut res = rest;
```
197g: `mem::replace(&mut self.rest, rest)` seems more fitting for this and the next line.

src/cskvp.rs
```diff
@@ -138,3 +137,129 @@ impl<'a> Drop for Cskvp<'a> {
...
+    fn next_quoted(&mut self) -> Cow<'a, str> {
+        assert!(self.rest.trim_start().starts_with('"'));
+
+        match &self.rest {
+            Cow::Borrowed(rest) => match quoted_string::parse::<TestSpec>(rest) {
+                // TODO: error handling
+                Err(e) => panic!("Invalid double-quoted string: {:?}", e),
+                Ok(parsed) => {
+                    self.rest = Cow::Borrowed(parsed.tail);
+                    quoted_string::to_content::<TestSpec>(parsed.quoted_string).unwrap()
```
197g, line 167: Not sure if this should rely on the impl advertised as [`an example`](https://docs.rs/quoted-string/0.6.0/quoted_string/test_utils/struct.TestSpec.html) in the crate. And the error case could happen so unwrapp seems badly handled (maybe it can't but what about `UnquoteableCharQuoted` for example).

197g (changes requested):

src/cskvp.rs
oberien: If I'm going to write my own Spec, I could just go ahead and write the whole parsing myself. It's not that hard, and would only need a single pass for parsing and unescaping instead of two passes in the library.

Regarding errors, the documentation of `to_content` states that it'll error if the string isn't double-quoted. But we already know that it's double-quoted, because it's the result of `parse`. I'm wondering why the library didn't make the `Parsed` struct's field private with accessor methods, and accept the Parsed struct in the `to_content` function.

In general I'm not too happy with the library, but decided to use it anyways, because that way we don't need to implement the parsing manually, reducing sources of bugs.

197g (approved):

oberien:
@HeroicKatora What is your verdict regarding https://github.com/oberien/pundoc/issues/19#issuecomment-468943142 ? Good enough for now? Should we close #19?

197g:
I'm definitely fine with double escaped quotes for introducing quotes within quoted strings. Anything more will be questionable. This is of course more natural for a programmer, where double escaping is common when writing regex strings for example, but I'd rather it be consistent with markdown (and hope for a raw string syntax in that at some point).

PR #31, opened 2019-03-13
Rename pundoc to heradoc

### LINKED ISSUE #11: Project Name

oberien:
Fix #11

When you approve this PR, I'll rename the repository before merging this PR to make all links in this PR valid.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -350,6 +350,35 @@ dependencies = [
 ]
 
 [[package]]
+name = "heradoc"
+version = "0.1.0"
+dependencies = [
+ "boolinator 2.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "env_logger 0.5.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "isolang 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "mime 0.3.12 (registry+https://github.com/rust-lang/crates.io-index)",
+ "pulldown-cmark 0.2.0 (git+https://github.com/raphlinus/pulldown-cmark?rev=07863e9e54cfcb108693ca1ab64967b5ad7bd952)",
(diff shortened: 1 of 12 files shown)


197g (approved):

oberien:
Friendly reminder to update your upstream URLs :)

PR #32, opened 2019-03-13
Allow input files in the config to contain links

oberien:
Resolve referenced input files in config both to input
file dir and current working dir.

For a usage example see https://github.com/oberien/heradoc/blob/examples/examples/thesis/thesis.md

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -12,10 +12,12 @@ use structopt::StructOpt;
 use serde::{Deserialize, Deserializer, de};
 use tempdir::TempDir;
 use isolang::Language;
+use url::Url;
 
 mod geometry;
 
 use self::geometry::Geometry;
+use crate::resolve::remote::Remote;
 
 // TODO: VecOrSingle to allow `foo = "bar"` instead of `foo = ["bar"]` for single values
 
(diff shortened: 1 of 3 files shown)


197g:
This looks like a cool feature, and something I wanted some time ago as well, but the double temporary directory somehow seems overly complicated. Would it not be possible to defer the resolution of the images/supplementary files until after the instantiation of the config? I don't think anything relies on their immediate availabilty in particular and if it does, it probably shouldn't.

oberien:
There isn't a double temporary directory. There is only a single directory. The weird code is due to the difference between the `temp_dir` and `out_dir`. If the user provides an `out_dir`, we use that for build files. We use the `temp_dir` for temporarily downloaded files. If the user *doesn't* specify an `out_dir` though, we put all build files into the `temp_dir` as well and only move the final result back to the working directory.

I'm just using the `temp_dir` to download the files from the config, just as we do when downloading any content. The only problem is that if the user specifies an `out_dir`, we want to clean it before rebuilding in it. But if the `out_dir` is the `temp_dir`, we don't want to clean it, as (1) it is already clean as it's a newly created temporary directory, and (2) we've already downloaded the config input files into it.

Edit: There may be a better way to express this in the `main.rs` / `config/mod.rs`, but I don't really know how we should do that without introducing too much redundancy.

oberien:
> Would it not be possible to defer the resolution of the images/supplementary files until after the instantiation of the config?

Yes and no. Yes, it would be possible, but I'd like to verify the config while merging the three different configs. That way a config error can be triggered instantaneously. Also having everything related to config resolution in a single place feels better structure wise.

197g (approved):
Feature trumps form in that case. Especially with the future changes of merging the download directory that should improve this. Better errors are also desirable.

PR #47, opened 2019-03-14
Add unicode example

oberien:


197g:
Might be a good possibility to insert more tests in more exotic places:
* Show syntactical difference between inline code and code block context
* Metachars `#` in inline code, that was a problem once

oberien:
I implemented escaping all latex special chars correctly and added tests for it. See #53 for further information.

PR #48, opened 2019-03-14
Add math example

oberien:


197g:
Might be a good possibility to insert more math tests in more exotic places:
* Show latex meta characters like `#` and `{}` in inline code and code blocks (this overlaps with Unicode part)
* Could also insert a test for math within figure/math description

oberien:
> Could also insert a test for math within figure/math description

Actually we don't support that yet. I've opened #52. I think it'll be part of the rewrite where I want to remove the dependency from the frontend to the `PrimitiveGenerator`.

Issue #53, opened 2019-03-15
Correct handling of latex special chars


oberien:
Latex has [10 special chars](http://www.emerson.emory.edu/services/latex/latex_155.html): `# $ % & ~ _ ^ \ { }` Currently (after #47) we translate them as text symbol (instead of as control character), whenever they are encountered (except `\`, which we're checking for a possible latex command and only escape if it is definitely not a command). We may want to handle them context-awarely to make inline latex work as expected.

Another idea is to have a special way of defining inline latex, which is translated differently, like
````markdown
```inlinelatex
\whatevercommandyouwanttouse
```
````

That way we can assume that no inline-latex will be present and just translate everything. If we go this way we could just drop special blocks in all other backends and warn the user. The user can define different block for each backend, in wihch case we don't output a warning:

````
```inlinelatex
\whatever
```
```inlinehtml
<whatever />
```
````
(Ok, bad example, because inline-html should already be supported according to CommonMark anyways. We could think about disallowing it anyways and forcing this method onto our users instead.)

**Comment by 197g** [comment]
One problem with code blocks is that I typically may want to *mix* latex content with heradoc to get the best of both worlds. For example to open some custom `minipage` environment. Surrounding everything with triple quoted code blocks looks extremely clunky. Maybe we can leverage inline code blocks for a similar effect:

```
`\begin{minipage}{0.5\line}`
**Here** I would still want the full improvements from heradoc.
`\end{minipage}`
```

This could be complementary to the full code blocks! Basically establishing again that inline code blocks should have the same power as full ones, though it would require some special syntax considerations as it is less obvious than `$ math^2`

**Comment by 197g** [comment]
Related, I think `[include la.tex]` (or `![](la.tex)`) should eventually behave the same as an `inlinelatex` code block as well (I'm fine with doing these in steps). That includes potentially disallowing it without a declared latex feature.

**Comment by oberien** [comment]
The idea is that you are free to mix latex blocks / environments with heradoc (and that's what I'm using it for):
````
```inlinelatex
\begin{landscape}
```
text foo bar
```inlinelatex
\end{landscape}
```
````

I do agree that it would be nice to have some sort of inline-syntax for latex similar to inline math. I don't know which character to promote to a metachar, but my first though would be a backslash instead of a dollar (like math) followed by a space: `\ \foo`.

197g:
[...]

PR #55, opened 2019-03-18
Inline Latex?

oberien:


src/frontend/mod.rs
```diff
@@ -146,6 +146,32 @@ impl<'a, B: Backend<'a>> Frontend<'a, B> {
         }
     }
 
+    fn convert_text(&mut self, mut text: Cow<'a, str>) {
+        // test if the text contains a latex command
+        lazy_static! {
+            // latex actually isn't really parsable, so this is only a very rough estimate of common commands
+            // Expect: \commandname<arg>[arg]{arg}{arg}...
+            static ref LATEX_COMMAND: Regex = Regex::new(r"\\[a-zA-Z]+(?:$|<.*?>|\[.*?\]|\{.*?\})").unwrap();
```
197g, line 154: What about the regex at https://github.com/oberien/heradoc/blob/548612591ffd0ac58bd9a56e042d85da34e4ac9f/src/backend/latex/simple.rs#L32? I'm a big confused about the purpose of both, and maybe an explanation in the docu might be good for visually parsing both and mentioning their individual limitations.

src/frontend/mod.rs
```diff
@@ -146,6 +146,32 @@ impl<'a, B: Backend<'a>> Frontend<'a, B> {
...
+        }
+        if let Some(m) = LATEX_COMMAND.find(&text) {
+            let (start, end, len) = (m.start(), m.end(), m.as_str().len());
+            if start != 0 {
+                let (l, r) = text.split_at(start);
+                text = r;
+                self.buffer.push_back(Event::Text(l));
+            }
+            if len != text.len() {
+                let (l, r) = text.split_at(end);
```
197g, line 164: Pretty sure this is the wrong index, `text` is the tail beginning from the regex match while `end` refers to the index in the original text. Avoiding split for empty slices is maybe not any faster either, surely not for slice and probably not for `String`, so rather something simple.

```
let (head, from_match) = text.split_at(start);
let (latex, tail) = tex.split_at(len);
if !head.is_empty() {
    self.buffer.push_back(Event::Text(head))
}
self.buffer.push_back(Event::Latex(latex));
// tail omitted, see next comment
```

src/frontend/mod.rs
```diff
@@ -146,6 +146,32 @@ impl<'a, B: Backend<'a>> Frontend<'a, B> {
...
+                self.buffer.push_back(Event::Latex(l));
+                self.buffer.push_back(Event::Text(r));
```
197g, line 166: Wait, what happens when there is a second latex command in the same piece of text?

197g (changes requested):

src/frontend/mod.rs
oberien: The regex in the text-generator is removed in this PR. It was used to detect possible latex command to prevent escaping `\\` before the command. It was pretty sophisticated and should detect most beginnings of latex commands. It would probably have a lot of false positives.

The new regex used in the frontend is a very rough estimate and won't catch most special command syntaxes. It should catch any command of the form `\name<args>[args]{args}`, but will probably have some false negatives.

src/frontend/mod.rs
oberien: Good point, I should probably move this into a loop.

oberien:
What is your comment on #53 ? If we decide to go for that approach, we don't need this regex at all.

oberien:
Actually I just realized that this idea isn't as simple as I tried to solve it in this PR. Currently I miss that latex commands can also be ended with just a simple space. If I include that into the regex, everything breaks, because we can only do that transformation context-sensitively. For example `println!("foo\nbar baz")` in code shouldn't be parsed as `Text("println!(\"foo\")), Latex("\nbar"), Text(" baz\"))`. But in the frontend we're only performing contextfree transformations. As such, we can't do this sophisticatedly in the frontend.

Therefore I'm going to close this PR and instead implement the approach layed out in #53.

PR #56, opened 2019-03-20
Inline latex via #53

### LINKED ISSUE #53: Correct handling of latex special chars

oberien:
Let's try this again :)

Fix #53


197g (approved):
No harm in doing the inline code block addition in another pull request, this is definitely nicer than regex.

PR #60, opened 2019-03-21
Update pulldown_cmark dependency

### LINKED ISSUE #30: CollapsedUnknown isn't produced by [#foo][]
This is a bug in pulldown-cmark as it seams. It's probably not worth opening an issue about it with the new_alg branch being merged in raphlinus/pulldown-cmark#200. We should retest this on the new alg.

oberien:
Also include all new features of pulldown-cmark:
* Tasklists
* Strikethrough

Fix #30


src/frontend/mod.rs
```diff
@@ -85,6 +87,7 @@ impl<'a, B: Backend<'a>> Frontend<'a, B> {
             })),
             CmarkEvent::SoftBreak => self.buffer.push_back(Event::SoftBreak),
             CmarkEvent::HardBreak => self.buffer.push_back(Event::HardBreak),
+            CmarkEvent::TaskListMarker(checked) => self.buffer.push_back(Event::TaskListMarker(checked)),
```
197g: `FootnoteReference` and `Pdf` event types each have their own struct encapsulation for their data, not sure if that would be cleaner to introduce here as well.

197g (approved):
Just a minor style inconsistency.

PR #63, opened 2019-03-25
Warnings and Rustfmt

oberien:
Fix #44

PR #65, opened 2019-03-27
Rid primitivegenerator and refactor Frontend link types

### LINKED ISSUE #64: Refactor frontend to not render stuff
Instead those Events should be returned as `Vec<Event>` and rendered in the Generator. This will also allows us to get rid of `PrimitiveGenerator`.

oberien:
Depends on #63 
Fix #64


oberien:
@HeroicKatora I rebased this into master, so the diff is more readable.

197g (approved):
Had I guessed from the title that this was mostly about also being to have different link types already in the frontend I would have done this sooner :sweat_smile: (That was is intentionally included in that PR, right?)

oberien:
That was a side-effect. If we can't render stuff anymore, we need to handle it differently. Previously we just rendered the content of a link. But now we can't anymore, because the frontend is just a "stupid" element in the pipeline. Beforehand it was able to access elements later in the pipeline (namely the `PrimitiveGenerator`).
Thus, instead of rendering the content, the whole link event infrastructure required a rewrite.

PR #69, opened 2019-04-01
Error reporting, major refactors

### LINKED ISSUE #21: Fancy error messages
Related to #13 we should have some sort of `Context` object, which contains the bounds of the element that pulldown-cmark is currently parsing such that we can display fancy error messages.

oberien:
This got larger than expected. To make this easier I have refactored the Generator into an `Iter` module and to enforce the `StackElement::Context` (which previously needed to be put on the stack by the caller).

Even if we print diagnostics / errors, I try to recover from then by skipping over the events resulting in diagnostics and continuing with the next ones. No matter how many diagnostics got printed, I still try to produce a pdf.

There are three types of errors:
1. Internal logic errors: They just panic and unwind the program.
2. Non-recoverable expected errors (Fatal errors): Like when the latex output stream errors while writing to it. They are bubbled up to the `main` and printed there.
3. Diagnostics: We try to recover from them and continue generating.

For a testrun, execute `cargo run -- examples/errors.md`.

Fix #21 

Example of an ICE output (this particular one should probably be fixed :) ):

```rust
- bench.md:214208:1
       |
214208 | ![Representation example](img/representation-overview.png)
       | -- defined here
       |
error: cause: No such file or directory (os error 2)

error: internal compiler error: error converting url to path
note: please report this
note: in heradoc file None name Some(backtrace::backtrace::trace_unsynchronized::h4edc62c197f1da24) line None address Some(0x55682b1aca6d)
note: in heradoc file None name Some(heradoc::diagnostics::Diagnostics::bug::h93d19dd5adb6e56f) line None address Some(0x55682b1ea7e5)
note: in heradoc file None name Some(heradoc::resolve::Resolver::resolve::h9d40ca2a9539900f) line None address Some(0x55682b1b5718)
note: in heradoc file None name Some(heradoc::generator::iter::Iter::next::h7b763bfe69fe5de2) line None address Some(0x55682b24ff81)
note: in heradoc file None name Some(heradoc::generator::iter::Iter::peek::ha0e91824631af9c0) line None address Some(0x55682b25477a)
note: in heradoc file None name Some(heradoc::generator::Generator<B,W>::generate_body::ha9a7b0303c2dce06) line None address Some(0x55682b23c948)
note: in heradoc file None name Some(heradoc::generator::Generator<B,W>::generate::h6c7617fc6ab18829) line None address Some(0x55682b233be5)
note: in heradoc file None name Some(heradoc::backend::generate::ha0f89092eff9bc78) line None address Some(0x55682b1f41e4)
note: in heradoc file None name Some(heradoc::main::h5c534d9e94344157) line None address Some(0x55682b1b8140)
note: in heradoc file None name Some(std::rt::lang_start::{{closure}}::h95a941aaf9dbbd15) line None address Some(0x55682b1a4c02)
note: in heradoc file Some("src/libstd/rt.rs") name Some("{{closure}}") line Some(49) address Some(0x55682b54bd92)
note: in heradoc file Some("src/libstd/panicking.rs") name Some("do_call<closure,i32>") line Some(293) address Some(0x55682b54bd92)
note: in heradoc file Some("src/libpanic_unwind/lib.rs") name Some("__rust_maybe_catch_panic") line Some(87) address Some(0x55682b555be9)
note: in heradoc file Some("src/libstd/panicking.rs") name Some("try<i32,closure>") line Some(272) address Some(0x55682b54c95c)
note: in heradoc file Some("src/libstd/panic.rs") name Some("catch_unwind<closure,i32>") line Some(388) address Some(0x55682b54c95c)
note: in heradoc file Some("src/libstd/rt.rs") name Some("lang_start_internal") line Some(48) address Some(0x55682b54c95c)
note: in heradoc file None name Some("main") line None address Some(0x55682b1bc281)
note: in heradoc file None name Some("__libc_start_main") line None address Some(0x7f17ad25b222)
note: in heradoc file None name Some("_start") line None address Some(0x55682b18719d)
note: in heradoc file None name None line None address None
```

### DIFF
--- a/src/cskvp.rs
+++ b/src/cskvp.rs
@@ -9,15 +9,17 @@ use single::{self, Single};
 use crate::ext::{CowExt, VecExt};
 use crate::diagnostics::Diagnostics;
 
+struct Diagnostic;
+
 #[derive(Debug)]
 pub struct Cskvp<'a> {
     diagnostics: Option<Diagnostics<'a>>,
     range: Range<usize>,
-    label: Option<Cow<'a, str>>,
-    caption: Option<Cow<'a, str>>,
-    figure: Option<bool>,
-    single: Vec<Cow<'a, str>>,
-    double: HashMap<Cow<'a, str>, Cow<'a, str>>,
+    label: Option<(Cow<'a, str>, Range<usize>)>,
+    caption: Option<(Cow<'a, str>, Range<usize>)>,
+    figure: Option<(bool, Range<usize>)>,
+    single: Vec<(Cow<'a, str>, Range<usize>)>,
+    double: HashMap<Cow<'a, str>, (Cow<'a, str>, Range<usize>)>,
 }
 
 impl<'a> Default for Cskvp<'a> {
@@ -35,56 +37,63 @@ impl<'a> Default for Cskvp<'a> {
 }
 
 impl<'a> Cskvp<'a> {
-    pub fn new(s: Cow<'a, str>, range: Range<usize>, diagnostics: Diagnostics<'a>) -> Cskvp<'a> {
-        let parser = Parser::new(s);
+    pub fn new(s: Cow<'a, str>, range: Range<usize>, content_range: Range<usize>, mut diagnostics: Diagnostics<'a>) -> Cskvp<'a> {
+        let mut parser = Parser::new(s, content_range);
 
         let mut single = Vec::new();
         let mut double = HashMap::new();
-        let mut label = None;
-        for value in parser {
+        let mut label: Option<(_, Range<usize>)> = None;
+        while let Ok(Some((value, range))) = parser.next(&mut diagnostics) {
             match value {
                 Value::Double(key, value) => {
-                    double.insert(key, value);
+                    double.insert(key, (value, range));
                 },
                 Value::Single(mut value) => {
                     if value.starts_with('#') {
                         if label.is_some() {
-                            // TODO: warn
-                            println!(
-                                "Found two labels, taking the last: {} and {}",
-                                label.as_ref().unwrap(),
-                                &value[1..]
-                            );
+                            diagnostics
+                                .warning("found two labels")
+                                .with_info_section(&label.as_ref().unwrap().1, "first label defined here")
+                                .with_info_section(&range, "second label defined here")
+                                .note("using the last")
+                                .emit();
                         }
                         value.truncate_start(1);
-                        label = Some(value);
+                        label = Some((value, range));
                     } else {
-                        single.push(value);
+                        single.push((value, range));
                     }
                 },
             }
         }
 
-        let figure_double = double.remove("figure").and_then(|s| match s.parse() {
-            Ok(val) => Some(val),
+        let figure_double = double.remove("figure").and_then(|(s, range)| match s.parse() {
+            Ok(val) => Some((val, range)),
             Err(_) => {
-                // TODO: warn
-                println!("cannot parse `figure={}`, only `true` and `false` are allowed", s);
+                diagnostics
+                    .error("cannot parse figure value")
+                    .with_error_section(&range, "defined here")
+                    .note("only `true` and `false` are allowed")
+                    .emit();
                 None
             },
         });
-        let figure = single.remove_element(&"figure").map(|_| true);
-        let nofigure = single.remove_element(&"nofigure").map(|_| false);
+        let figure = single.remove_element(&"figure").map(|(_, range)| (true, range));
+        let nofigure = single.remove_element(&"nofigure").map(|(_, range)| (false, range));
 
-        let figure = match [figure_double, figure, nofigure].iter().cloned().flatten().single() {
+        let figures = [figure_double, figure, nofigure];
+        let figure = match figures.iter().cloned().flatten().single() {
             Ok(val) => Some(val),
             Err(single::Error::NoElements) => None,
             Err(single::Error::MultipleElements) => {
-                // TODO: warn
-                println!(
-                    "only one of `figure=true`, `figure=false`, `figure` and `nofigure` allowed, \
-                     found multiple"
-                );
+                let mut diag = Some(diagnostics
+                    .error("found multiple figure specifiers"));
+                for (_, range) in figures.iter().cloned().flatten() {
+                    diag = Some(diag.take().unwrap().with_info_section(&range, "one defined here"));
+                }
+                diag.unwrap()
+                    .note("only one of `figure=true`, `figure=false`, `figure` and `nofigure` is allowed")
+                    .emit();
                 None
             },
         };
@@ -108,19 +117,19 @@ impl<'a> Cskvp<'a> {
         self.label.is_some()
     }
 
-    pub fn take_label(&mut self) -> Option<Cow<'a, str>> {
+    pub fn take_label(&mut self) -> Option<(Cow<'a, str>, Range<usize>)> {
         self.label.take()
     }
 
-    pub fn take_figure(&mut self) -> Option<bool> {
+    pub fn take_figure(&mut self) -> Option<(bool, Range<usize>)> {
         self.figure.take()
     }
 
-    pub fn take_caption(&mut self) -> Option<Cow<'a, str>> {
+    pub fn take_caption(&mut self) -> Option<(Cow<'a, str>, Range<usize>)> {
         self.caption.take()
     }
 
-    pub fn take_double(&mut self, key: &str) -> Option<Cow<'a, str>> {
+    pub fn take_double(&mut self, key: &str) -> Option<(Cow<'a, str>, Range<usize>)> {
         self.double.remove(key)
     }
 
@@ -144,27 +153,41 @@ impl<'a> Drop for Cskvp<'a> {
             .as_mut()
             .map(|d| d
                 .warning("in element config")
-                .with_section(range, "")
+                .with_error_section(range, "")
             );
-        // TODO: warn
-        if let Some(label) = self.label.as_ref() {
-            diag = diag.map(|d| d.warning(format!("label ignored: {}", label)));
+        if let Some((label, range)) = self.label.as_ref() {
+            diag = diag.map(|d| {
+                d.warning(format!("label ignored: {}", label))
+                    .with_info_section(range, "label defined here")
+            });
             has_warning = true;
         }
-        if let Some(figure) = self.figure {
-            diag = diag.map(|d| d.warning(format!("figure config ignored: {}", figure)));
+        if let Some((figure, range)) = self.figure.as_ref() {
+            diag = diag.map(|d| {
+                d.warning(format!("figure config ignored: {}", figure))
+                    .with_info_section(range, "figure config defined here")
+            });
             has_warning = true;
         }
-        if let Some(caption) = self.caption.as_ref() {
-            diag = diag.map(|d| d.warning(format!("caption ignored: {}", caption)));
+        if let Some((caption, range)) = self.caption.as_ref() {
+            diag = diag.map(|d| {
+                d.warning(format!("caption ignored: {}", caption))
+                    .with_info_section(range, "caption defined here")
+            });
             has_warning = true;
         }
-        for (k, v) in self.double.drain() {
-            diag = diag.map(|d| d.warning(format!("unknown attribute `{}={}`", k, v)));
+        for (k, (v, range)) in self.double.drain() {
+            diag = diag.map(|d| {
+                d.warning(format!("unknown attribute `{}={}`", k, v))
+                    .with_info_section(&range, "attribute defined here")
+            });
             has_warning = true;
         }
-        for attr in self.single.drain(..) {
-            diag = diag.map(|d| d.warning(format!("unknown attribute `{}`", attr)));
+        for (attr, range) in self.single.drain(..) {
+            diag = diag.map(|d| {
+                d.warning(format!("unknown attribute `{}`", attr))
+                    .with_info_section(&range, "attribute defined here")
+            });
             has_warning = true;
         }
 
@@ -178,6 +201,7 @@ impl<'a> Drop for Cskvp<'a> {
 
 struct Parser<'a> {
     rest: Cow<'a, str>,
+    range: Range<usize>,
 }
 
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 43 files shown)


197g:
Sweet. I'll try to get to it, as a first impression there are some opportunities for some cleanup. For example extracting `Range<usize>` into a newtype wrapper to make the import cleaner, add helper methods for interfacing with `diagnostics()..with_info_section(..)` and possibly adding additional information to the source location (intermediate files?) without requiring more rewrites.

197g:
@oberien Also please push your changes to `codespan-reporting`, `cargo` can't seem to find the respective commits (or are they https://github.com/oberien/codespan/commit/dc7b13739165fc58721415ce3120d05cf117ca81 after a rebase&force-push from upstream?).

oberien:
> cargo can't seem to find the respective commits 

Oops, after brendanzab/codespan#55 was merged, I rebased brendanzab/codespan#56 onto the new master, but forgot to update the revision hash in `Cargo.toml`. Should be fixed now.

> there are some opportunities for some cleanup

Oh absolutely, most likely. I wrote this in 3 days, and there's probably some stuff which I just hacked in without properly integrating it.

> extracting `Range<usize>` into a newtype wrapper to make the import cleaner

I'm not sure if that actually simplifies a lot of stuff, because especially in the frontend / cskvp we need to perform a lot of range calculations. I don't think that a newtype would help here, in fact it might even lower the user experience (with us being the user).
There is the `codespan::Span`, but I decided against it, because it's a hassle to work with. That's why I only convert to it last second in the `Diagnostics`.
I do see some advantages of redefining `Range<usize>` as `SourceRange`, which we can make `Copy` (and *not* implement Iterator on). That might clean somes tuff up (especially all the `clone`s throughout the code).

> add helper methods for interfacing with `diagnostics()..with_info_section(..)`

I'm not sure I'm following. I find the diagnostics API pretty nice actually :) (also having rewritten it thrice to better fit the needs)

> adding additional information to the source location

Again I'm not sure I'm following.

Anyways, I'm going to wait on your review, that might clear things up :)

src/cskvp.rs
```diff
@@ -108,27 +155,59 @@ impl<'a> Cskvp<'a> {
...
     fn drop(&mut self) {
-        // TODO: warn
-        if let Some(label) = self.label.as_ref() {
-            println!("label ignored: {}", label);
+        let mut has_warning = false;
+        let range = &self.range;
+        let mut diag = self
+            .diagnostics
+            .as_mut()
+            .map(|d| d.warning("in element config").with_error_section(range, ""));
```
197g: This shouldn't be its own diagnostic, rather an information section. Especially for a single warning, the warning text is rather confusing otherwise.

src/diagnostics.rs
```diff
@@ -0,0 +1,204 @@
...
+    code: Option<String>,
+    labels: Vec<Label>,
+}
+
+impl<'a: 'b, 'b> DiagnosticBuilder<'a, 'b> {
+    pub fn emit(self) {
+        let Self { file_map, out, mut diagnostics, severity, message, code, labels } = self;
+        diagnostics.push(Diagnostic { severity, message, code, labels });
+
+        // ignore output errors, because where would we log them anyway?!
```
197g, line 124: Yes, where would we log them. But wouldn't that mean that it's more sensible to `panic!()` instead of, say, continue trying to write to a full disk or buffer.

src/diagnostics.rs
```diff
@@ -0,0 +1,204 @@
...
+        for diagnostic in diagnostics {
+            let _ = codespan_reporting::emit_single(&mut *out, file_map, &diagnostic);
+        }
+        writeln!(out);
```
197g: Should use `let _ = ` to make intent clear. Note that this is currently suppressed due to `allow(unused)`.

src/diagnostics.rs
```diff
@@ -0,0 +1,204 @@
+#![allow(unused)]
```
197g: If the intent is to allow unused function interface, rather use `allow(dead_code)` as `unused` suppressed several other warnings, such as unused `must_use` and unused arguments.

src/diagnostics.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn clone(&self) -> Self {
+        Diagnostics { file_map: self.file_map.clone(), out: Diagnostics::create_out_stream() }
+    }
+}
+
+impl<'a> fmt::Debug for Diagnostics<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Diagnostics")
+            .field("file_map", &self.file_map)
+            .field("out", &"Rc(StandardStream)")
```
197g: I find this confusing as `out` is not an `Rc` itself.

src/frontend/concat.rs
```diff
@@ -1,30 +1,36 @@
...
+            Some((evt, range)) => return Some((evt, range)),
         };
 
-        while let Some(Event::Text(_)) = self.0.peek() {
-            let next = match self.0.next() {
-                Some(Event::Text(t)) => t,
+        while let Some(Event::Text(_)) = self.0.peek().map(|t| &t.0) {
+            let (evt, r) = self.0.next().unwrap();
+            // TODO: why are both variants needed?
+            assert!(range.end == r.start || range.end + 1 == r.start);
```
197g, line 29: This is indeed confusing.

src/diagnostics.rs
```diff
@@ -0,0 +1,204 @@
...
+use backtrace::Backtrace;
+use codespan::{ByteOffset, FileMap, FileName, Span};
+use codespan_reporting::termcolor::{ColorChoice, StandardStream};
+use codespan_reporting::{Diagnostic, Label, LabelStyle, Severity};
+use url::Url;
+
+use crate::resolve::Context;
+
+pub struct Diagnostics<'a> {
+    file_map: Rc<FileMap<&'a str>>,
```
197g: The `FileMap` is shared with an `Rc` but the contained strings by reference. Avoiding lifetimes entirely with, e.g. `Rc<str>`, seems more appropriate to hold the original source string in this case. Also, the `Rc`-ness of `file_map` is not used here as `DiagnosticsBuilder<_, _>` instead contains a reference. This seems to me as if motivated purely for a `.clone()` optimization. In that case, I wonder if it should not be the `Diagnostics` that should be wrapped in `Rc` or some other choice by the user, and `out` also not be configured herein but separately.

Quite confusing like this.

src/diagnostics.rs
```diff
@@ -0,0 +1,204 @@
...
+        self.diagnostic(Severity::Help, message.into())
+    }
+
+    pub fn with_error_code(mut self, code: String) -> Self {
+        self.code = Some(code);
+        self
+    }
+
+    fn with_section<S: Into<String>>(
+        mut self, style: LabelStyle, range: &Range<usize>, message: S,
```
197g: Taking `Range` by reference?

src/frontend/event.rs
```diff
@@ -96,15 +100,15 @@ pub enum Tag<'a> {
 
 #[derive(Debug, Clone)]
 pub struct Header<'a> {
-    pub label: Cow<'a, str>,
+    pub label: (Cow<'a, str>, Range<usize>),
```
197g: Repeatedly adding all the `(_, Range<usize>)` would have gotten me to write a type alias. Also makes it easier to change `Range<usize>` to something else if need be. For example if `Range<usize>` should be changed to something else or if we need the an identifier for the source file of this event etc .

src/generator/iter.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::error::{Error, FatalResult, Result};
+use crate::frontend::{Event as FeEvent, EventKind as FeEventKind, Frontend, Include as FeInclude};
+use crate::generator::event::{Event, Image, Pdf};
+use crate::generator::Generator;
+use crate::resolve::{Context, Include};
+
+pub struct Iter<'a> {
+    frontend: Fuse<Frontend<'a>>,
+    peek: Option<(Event<'a>, Range<usize>, FeEventKind)>,
+    last_kind: FeEventKind,
```
197g, line 21: `last_kind` doesn't quite tell me what this is used for, and what exactly its invariant is.

src/generator/iter.rs
```diff
@@ -0,0 +1,182 @@
...
+            | FeEventKind::Label
+            | FeEventKind::SoftBreak
+            | FeEventKind::HardBreak
+            | FeEventKind::TaskListMarker
+            | FeEventKind::Command
+            | FeEventKind::ResolveInclude => return Ok(()),
+        }
+        let mut depth = 0;
+        loop {
+            let evt = if let Some((peek, _, _)) = self.peek.take() {
```
197g: Shouldn't `self.next()` also `take` the correct event?

src/generator/iter.rs
```diff
@@ -0,0 +1,182 @@
...
+        gen: &mut Generator<'a, impl Backend<'a>, impl Write>,
+    ) -> Result<Event<'a>> {
+        match include {
+            Include::Command(command) => Ok(command.into()),
+            Include::Markdown(path, context) => {
+                let markdown = fs::read_to_string(&path).map_err(|err| {
+                    gen.diagnostics()
+                        .error("error reading markdown include file")
+                        .with_error_section(&range, "in this include")
+                        .error(format!("cause: {}", err))
```
197g, line 139: Include the resolved path: `info(format!("{}", &path))`. Several other tools omit these, making it very unclear why some `NotFound` actually occurs.. Also, this currently doesn't occur as it crashes during path resolution for an unfound file. What?

src/generator/mod.rs
```diff
@@ -183,58 +189,28 @@ impl<'a, B: Backend<'a>, W: Write> Generator<'a, B, W> {
...
+        self.stack
+            .iter_mut()
+            .rev()
+            .filter_map(|state| match state {
+                StackElement::Context(_, diagnostics) => Some(diagnostics),
                 _ => None,
             })
-            .expect("no Context???");
-        self.resolver.resolve(context, url)
+            .next()
```
197g: Extract this chain into a `fn top_context(&mut self) -> &mut (_, _)` to
* make the invariant clear
* reuse it here and in `resolve`

src/generator/stack.rs
```diff
@@ -14,7 +15,7 @@ impl<'a: 'b, 'b, D: Backend<'a> + 'b, W: Write> Stack<'a, 'b, D, W> {
     }
 
     pub fn iter(&self) -> impl Iterator<Item = &StackElement<'a, D>> {
-        self.stack.iter()
+        self.stack.iter().rev()
```
197g, line 18: ^^

src/resolve/source.rs
```diff
@@ -66,52 +97,70 @@ impl Source {
...
                     Ok(downloaded) => downloaded,
-                    Err(RemoteError::Io(io)) => return Err(io),
-                    // TODO: proper error handling with failure
-                    Err(RemoteError::Request(_req)) => {
-                        return Err(io::ErrorKind::ConnectionAborted.into());
+                    Err(RemoteError::Io(err)) => {
+                        diagnostics
+                            .error("error writing downloaded content to cache")
+                            .with_error_section(&range, "trying to download this")
+                            .error(format!("cause: {}", err))
```
197g, line 200: Also include the path the cached file should be written to, this would need to be propagated to the `Error` from `Remote`. It can not be resolve before, as the content type depends on the server response and possible cache presence.

This can be opened as a separate issue if you want.

197g (changes requested):

src/diagnostics.rs
oberien: Oops, not anymore, that is. Somewhere in between it was, before I refactored it out again.

src/frontend/concat.rs
oberien: This is actually just wrong, and I'll inspect it more soon-ish. (e.g. running it on our current `README.md`, that assertion fails) I just don't want to make this PR even larger, as such I left this as something to do after it lands.

197g: Seems reasonable, a single skipped character index that does not affect output.

197g:
Forgot the major comment on the PR: I really, really like having diagnostics!

src/error.rs
```diff
@@ -0,0 +1,43 @@
+use std::io;
+use std::result;
+
+#[must_use]
+pub type Result<T> = result::Result<T, Error>;
+#[must_use]
+pub type FatalResult<T> = result::Result<T, Fatal>;
+
+#[must_use]
+pub enum Fatal {
```
197g, line 12: Missing `#[derive(Debug)]` for unwrap etc.

src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl From<io::Error> for Fatal {
+    fn from(err: io::Error) -> Self {
+        Fatal::Output(err)
+    }
+}
+
+#[must_use]
+pub enum Error {
```
197g, line 53: Also needs `#[derive(Debug)]` at least.

src/cskvp.rs
```diff
@@ -177,65 +298,64 @@ impl<'a> Parser<'a> {
...
-        } else {
-            // TODO: error handling
-            assert!(delim == Some(',') || delim == None, "invalid comma seperated value");
-            Some(Value::Single(key))
-        }
-    }
-}
-
 #[cfg(test)]
 mod test {
```
197g, line 346: The test suite doesn't compile right now. Problably has been updated before some interface changes and then not run again? Some method returns in several `assert_eq` look like they are now `Result` but had inner unwraps before.

197g (changes requested):
Makes sure the tests work neatly

src/diagnostics.rs
oberien: tldr: While trying to justify what I did, I think the best solution in the long run is probably to have `Arc<Diagnostic<'a>>` with
```rust
struct Diagnostic<'a> {
   file_map: FileMap<&'a str>,
   out: Arc<Mutex<StandardStream>>,
}
```

The current design stems from older iterations, in which using `Rc` or even `Rc<RefCell>` wouldn't easily have possible. Given the current structure, I guess the best approach is to already go for `Arc<Diagnostics<'a>>` and `Arc<Mutex<StandardStream>>` with #71 and #73 (parallelization) in mind.

I'm unsure if I should perform that major refactor now already, or if we do it at a later point after this is merged.

---

There are a multitude of issues at hand here, which resulted in me deciding for this solution. I thought about several solutions here, but this one seemed the least intrusive.

Let's start with the lifetime `'a`. The whole codebase is tainted with `'a`, because it's the lifetime of all input files' content. This stems from us wanting to use a single instance of a Backend for multiple input files. As such all input files, even the ones included at a later point, need the same lifetime `'a`. To allow that, the `Generator` is passed a `&'a Arena`, in which we allocate all file content `String`s. This gives us a `&'a str`, which we then use for everything.
The `codespan` crate expects its `CodeMap` to hold all file contents (as it only accepted `String` before brendanzab/codespan#55). In fact, the `emit` function only accepted a `CodeMap` and not a single `FileMap`. This doesn't fit our model and data layout. Thus I opened brendanzab/codespan#56 as well, which allows us to emit diagnostics given just the `FileMap`. This fits our use-case, as we are basically performing a recursive-descend for includes. As such there is no harm in having `FileMap<&'a str>`. In fact using `Rc<str>` instead of `&'a str` might result in problems with pulldown_cmark borrowing from the original `str`.
(Thinking about this I think instead of a Backend we could also accept a `BuildBackend` implementation, which allows us to create a new backend for each include, which can be generic over a different lifetime. That way we don't need the arena and also don't need to hold the `String` content after we're finished processing that file. If this is something worth investigation in your opinion @HeroicKatora, we should open an issue to keep it in mind. This might especially help with parallelization.)

To output a diagnostic, we need an output and the current `FileMap`. We need to be able to output diagnostics in the frontend, the generator, the backend and CodeGenUnits. We create a new frontend for each file, but we reuse the same generator for multiple files. The frontend is consumed by the events-iterator, which is passed as argument to functions (again due to the recursive-descend). This means that the generator doesn't easily have access to the `Diagnostics` of the frontend. Thus it needs its own instance of the diagnostics on the stack, which I moved into the existing `StackElement::Context`. This means that we need two different points where we need to access the `Diagnostics`, but one can't borrow from the other.
I agree with you that the clean way would probably be to wrap the `Diagnostics` itself in an `Rc` / `Arc`. I honestly don't know where the best place to create the `StandardStream` is, given during creation we must have access to the `Config`, and we should also probably synchronize access to the `StandardStream`, leaving us with `Arc<Mutex<StandardStream>>` in the concurrent scenario. That means we should probably create a single instance in the `Generator` and clone it for usage everywhere else.

src/diagnostics.rs
oberien: `Range` doesn't implement `Copy`. For easier usability I accept a reference here, instead of requiring all users to always call `.with_section(range.clone(), "foo")`. I guess as discussed in https://github.com/oberien/heradoc/pull/69#issuecomment-478477151 a `SourceRange` type could make sense to get rid of this kind of stuff by implementing `Copy` on it.

197g: Since the only place where you'd construct them by hand and thus really benefit from language semantics (`..`) is within the tests, that would be quite reasonable. The most immediate other benefit is being able to use it as an index parameter but the final usage of `Range` currently is constructing the argument to the diagnostics library.

src/diagnostics.rs
oberien: If we panic, that'll still try to write a message to the output, which is already gone. So the only sane way to handle this would be an `abort()`, but then the user might wonder why the program is just gone without any error message.

So there isn't really any sane way to handle this. This is something that just shouldn't happen :) I guess I should just panic here and let the `panic_handler` take care of the rest. Anyway, even this might be bad, because the panic handler writes stuff again.

src/diagnostics.rs
oberien: Yeah, allowed the wrong lint. Wanted `dead_code` actually.

src/frontend/concat.rs
oberien: I opened #76

src/generator/iter.rs
oberien: > Also, this currently doesn't occur as it crashes during path resolution for an unfound file.

That should be fixed with #75. After that this diagnostic should be fine.

src/cskvp.rs
oberien: Should I still do this or are you fixing this in #77?

197g: I think it should be fixed here.

oberien:
I think I fixed all your requests, so this PR should be ready for a second round of reviews. Sorry that this became so huge, we should definitely think about #80 in the long run to keep these kinds of PRs smaller. Also I don't really know how I would split up this PR into smaller ones.

197g:
I'm no longer sure if emitting diagnostics immediately into its own defined writer is optimal. If you run the tests from the `resolve-error` branch you currently get output printed even if successful because the stream opened in `Diagnostics` does not respect the redirection performed by the test harness and some tests have diagnostics even if they succeed. Maybe it would be better to buffer them into an intermediate and/or don't tie the construction of the output stream *into* the `Diagnostics` constructors.

And sorry that this appears so late in the progress and you literally changed hundreds of lines concerning exactly this :/

197g:
Alternatively, we could change the owner semantics of the current `DiagnosticsBuilder` aggregator and pass it by mutable reference in these cases, then delete the events before they would be printed or simply forget it since they only get printed by an explicit call to `emit`.

oberien:
I'll take a look into it after some sleep :) If we can improve it, I think it's worth it to investigate other possibilities. I was wondering as well why it printed out the diagnostics, considering that the rustc test framework captures it. I think that `StandardStream::stderr` will open a completely new `stderr` stream instead of using filedescriptor `2`. That maybe creates a new filedescriptor which rustc can't capture. I find this weird as I was expecting it to just wrap `std::io::Stderr`.

One idea I also had in the beginning was to collect all diagnostics in the `Diagnostics` and print them explicitly / on drop. I figured out that it was way easier to just directly print them.

Maybe we can just use a `termcolor::NoColor<std::io::Stderr>` or `termcolor::Ansi<std::io::Stderr> for tests?

197g:
> Maybe we can just use a termcolor::NoColor<std::io::Stderr> or `termcolor::Ansistd::io::Stderr for tests?

Something like that could also work, simply provide another constructor for tests. This is also more of a code-cleanup than a blocking concern for the PR as normal program experience is not itself affected for now. Should we ever get a case where we error but later in processing can completely 'unerror' or 'unwarn' then this becomes more important. But for now, just fixing it for tests is likely fine.

oberien:
Actually it seems like `termcolor::StandardStream` internally uses `std::io::Stderr`: <https://docs.rs/termcolor/1.0.4/src/termcolor/lib.rs.html#224-229>

I have no idea why it shows up in the tests.

oberien:
Actually it seems like [cargo test only captures stdout and not stderr](https://github.com/rust-lang/rust/issues/40298). So I guess we could just use `StandardStream::stdout` for unit tests.

oberien:
Apparently `cargo test` only captures stuff printed through `println` and `eprintln`?! Because even when I switch to `StandardStream::stdout`, it won't get captured. Will need more reading into, but this is pretty bad for `cargo test`…

oberien:
I think we can move this discussion to after this PR has landed. For the time being it's fine and it can always be refactored. I've opened #81 to track the diagnostics issue.

197g (approved):

oberien:
Thanks for the reviews!

Issue #73, opened 2019-04-03
Optimizations (ifneedbe)


oberien:
This issue is a collection of possible optimizations which might or might not be needed. Benchmarks can be seen in the comments.

- [ ] frontend: Use state-machine. Currently we basically perform a recursive-descend on the events. When we find an event with inner elements (`Event::Start`), we convert the whole event until the corresponding `Event::End` and write all converted events into a buffer. We then remove elements from the buffer until it is empty before parsing the next event.  
This could be replaced with a state-machine, which only converts a single event at a time instead of buffering stuff. This will be quite more complex though and might not be worth it considering performance gain versus mental overhead.
- [ ] everywhere: Don't use `dyn Write`. Currently we use `dyn Write` everywhere, because there might be a lot of different types involved, which can be written to (`Stdout`, `File`, `Vec<u8>` for intermediate caching in CodeGenUnits). We could create a custom enum wrapper instead, which might enable further optimizations (which are not possible via vtables).
- [ ] concurrency: Currently only a single thread is used. There are two ways we can achieve multithreading.
    * CompilationUnit multithreading: This should be done after #71. For every compilation unit (i.e. markdown file encountered), we start a new thread which compiles that file. After every unit has finished compiling, we merge them into a single file (just as we need to do for #71 anyways).
    * Internal multithreading: We find a way to multithread the compilation of a single compilation unit. I don't really have any idea how to achieve that yet, but if it's needed, we can take a closer look.
- [ ] String types: Currently we use `Cow<'a, str>`, but may want to be more careful / switch to the `SmortStr` of `pulldown-cmark` which can hold a character inline. We may also want to think about not using the `Concat` module, which may create large strings.
- [x] BufWriter: https://github.com/oberien/heradoc/issues/73#issuecomment-479427795 Required (#74)
- [ ] ThinLTO: https://github.com/oberien/heradoc/issues/73#issuecomment-479430628 Only worth it for actual release artifacts, not for benchmarking / testing.

Template
````
Benchmark $WHAT on 1300× [crdt-details.md] (96MB):

tldr: 

`perf stat ./target/release/heradoc bench.md -o bench.tex`
```

```

`perf record`
```

```

[crdt-details.md]: https://github.com/xi-editor/xi-editor/blob/564cbde392a2ba8d78a069bcf3dc6aa6a14c2aae/docs/docs/crdt-details.md
````

oberien:
[...]

PR #74, opened 2019-04-03
Use BufWriter for output

oberien:
As discussed in #73, this boost performance by 50%.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,3 +30,6 @@ quoted-string = "0.6.0"
 
 [profile.release]
 debug = true
+
+[profile.bench]
+debug = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,3 +27,6 @@ regex = "1.1.0"
 lazy_static = "1.2.0"
 single = "1.0.0"
 quoted-string = "0.6.0"
+
+[profile.release]
+debug = true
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -27,3 +27,6 @@ regex = "1.1.0"
 lazy_static = "1.2.0"
 single = "1.0.0"
 quoted-string = "0.6.0"
+
+[profile.release]
+debug = true
```
197g, line 32: Is there a way to utilize the [`bench` profile](https://doc.rust-lang.org/cargo/reference/manifest.html#the-profile-sections) instead? Seems like an easy mistake to have a larger than intended binary in the actual release.

197g (commented):
Performance win trumps having smaller binary size but I would like `debug` to only be enabled for builds that are used in profiling. But the argument could be made to profile in the release binaries only, this is only an issue if it's easy to avoid.

Cargo.toml
oberien: The idea behind `debug=true` in release builds is the `ICE` diagnostic, which accesses the backtrace to generate an output. Otherwise we'd be presented with a useless bug report in release builds.

197g: Oh yes, good point. This just turns out to also helping benchmarking in the same way.

PR #77, opened 2019-04-06
Expect failed canonicalization correctly

197g:
Canonicalization is required to ensure proper access rights and failure
to do so (for example from a missing file) should not be treated as an
internal implementation failure. Rather, it is most likely the input
file's fault from a malformed path.

An intermediate error type diagnoses different problems with path
normalization when resolving a relative file.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -12,6 +12,9 @@ pub type FatalResult<T> = result::Result<T, Fatal>;
 pub enum Fatal {
     /// Output file write error.
     Output(io::Error),
+
+    /// An unrecoverable internal error.
+    InternalError,
 }
 
 impl From<io::Error> for Fatal {
@@ -24,6 +27,7 @@ impl std::error::Error for Fatal {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match self {
             Fatal::Output(io) => Some(io),
+            Fatal::InteralCompilerError => None,
         }
     }
 }
@@ -32,6 +36,7 @@ impl fmt::Display for Fatal {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Fatal::Output(io) => write!(f, "output file write error: {}", io),
+            Fatal::InteralCompilerError => write!(f, "can not continue due to internal error"),
         }
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -13,8 +13,15 @@ pub enum Fatal {
     /// Output file write error.
     Output(io::Error),
 
-    /// An unrecoverable internal error.
-    InternalError,
+    /// An unrecoverable internal error (ICE).
+    ///
+    /// Used for assertions that are made but not completely or badly proven. For cases where this
+    /// can be related directly to an particular structure in an input file this error also
+    /// provides opportunities to log the relevant context in diagnostics as opposed to `unwrap` or
+    /// `expect` panicking.  For example this may occur due to an interface of a dependency being
+    /// more general than its usage herein but we can not expect full control over its
+    /// implementation.
+    InteralCompilerError,
 }
 
 impl From<io::Error> for Fatal {
--- a/src/resolve/mod.rs
+++ b/src/resolve/mod.rs
@@ -189,7 +189,6 @@ mod tests {
 
         assert_match!(main, Include::Markdown(path, _) if path == &dir.path().join("main.md"));
         assert_match!(sibling, Include::Image(path) if path == &dir.path().join("image.png"));
-        drop(dir);
     }
 
     #[test]
@@ -203,7 +202,6 @@ mod tests {
             .expect("Failed to resolve path in different domain");
 
         assert_eq!(toc, Include::Command(Command::Toc));
-        drop(dir);
     }
 
     #[test]
@@ -222,7 +220,6 @@ mod tests {
             .expect("Failed to download external document");
 
         assert_match!(external, Include::Markdown(_, Context::Remote(_)));
-        drop(dir);
     }
 
     #[test]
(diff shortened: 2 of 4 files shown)


src/resolve/mod.rs
```diff
@@ -222,4 +222,46 @@ mod tests {
...
+
+    #[test]
+    fn local_resolves_not_exist_not_internal_bug() {
+        let (dir, range, mut diagnostics) = prepare();
+        let resolver = Resolver::new(PathBuf::from("."), dir.path().join("download"));
+        let top = Context::LocalRelative(Path::new(dir.path()).canonicalize().unwrap());
+
+        let error = resolver
+            .resolve(&top, "this_file_does_not_exist.md", range, &mut diagnostics)
+            .expect_err("Can not resolve source for files that do not exist on disk");
```
oberien: I find this double negation confusing. Could you try rephrasing it?

src/error.rs
```diff
@@ -7,9 +7,13 @@ pub type Result<T> = result::Result<T, Error>;
 pub type FatalResult<T> = result::Result<T, Fatal>;
 
 #[must_use]
+#[derive(Debug)]
 pub enum Fatal {
     /// Output file write error.
     Output(io::Error),
+
+    /// An unrecoverable internal error.
+    InternalError,
```
oberien: I think `InternalCompilerError` is more clear here. Reading `InternalError` I was about to comment that it's not specific enough and should be split up into proper suberrors. But judging from its usage, I think ICE would make the most sense.

src/resolve/mod.rs
```diff
@@ -222,4 +222,46 @@ mod tests {
...
+
+        assert_match!(error, Error::Diagnostic);
+        drop(dir);
```
oberien: This shouldn't be needed, should it?

src/resolve/source.rs
```diff
@@ -198,11 +257,29 @@ fn to_include(
...
-    Ok(path)
+fn to_path<P: AsRef<Path>>(url: &Url, workdir: P) -> std::result::Result<PathBuf, PathError> {
+    let mut full_path = workdir.as_ref().to_path_buf();
+
+    url
+        .path_segments()
+        .ok_or(PathError::NoPath)?
+        .try_for_each(|segment| {
+            let mut components = Path::new(segment)
+                .components();
```
oberien, line 268: Is this really needed? `Url::path_segments` should already split at `/`, which should be enough.

src/resolve/source.rs
```diff
@@ -198,11 +257,29 @@ fn to_include(
...
-    env::set_current_dir(workdir)?;
-    let path = path.canonicalize()?;
-    env::set_current_dir(old_workdir)?;
-    Ok(path)
+fn to_path<P: AsRef<Path>>(url: &Url, workdir: P) -> std::result::Result<PathBuf, PathError> {
+    let mut full_path = workdir.as_ref().to_path_buf();
+
+    url
+        .path_segments()
+        .ok_or(PathError::NoPath)?
```
oberien, line 265: I can specify an include ``, which will result in an `InternalError`. I don't think that should be the case.

oberien (changes requested):

src/resolve/source.rs
197g: `path_segments` is split at `/` but could contain other OS specific separators or markers such as on Windows a literal `\` (? maybe not ?) , a file drive component `C:`, ... Basically, I wanted to make this as safe as possible and avoid any ambiguity.

src/resolve/source.rs
197g: What should that resolve to or error to? That sounds like a good test.

197g: Also, I had assumed the path to be `/` in that case. Seems like the assumption was wrong, then this may deserve its own variant.

oberien: Actually both `heradoc://document/` and `heradoc://document` have a path segment `""`: <https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=597179843b1f2dce6824cd357c06b45f>. So I guess we're fine here. I guess if `NoPath` can even happen here? If not, we should probably just `expect` it instead of possibly handling an impossible case.

197g: Good. It's not intended to ever occur, but I can't prove it never does, so that's why it would result in an `bug` message then. I'll add the `url` in question to the `Error` variant so that we know how this occured when a bug report is opened.

src/error.rs
197g: Maybe I should rename it to `AssertionFailed`? And then provide more documentation mentioning it exists for better integration into the diagnostics systems and to provide a proper, non-panicking error handling program flow where we rely on assumptions that we can't prove (e.g. because we don't have full control over them in the case of dependencies).

oberien: I think `InternalCompilerError` is fine here, because that's literally what it is. We print a diagnostic for one, so why shouldn't we return it as one? :)

oberien:
Other than renaming the fatal error to ICE, I think only a rebase is missing for a merge :)

src/error.rs
197g: I just find it a bit confusing labeling it a *compiler* error. That may be familiar for programmers but could be entirely confusing (in an already uncomfortable situation of a crash) to an other user demographic of which there are plenty others outside of programmers. But possibly fine?

oberien: I mean in the `Display` impl / `main` method we can still just have `"\n\nCan not continue due to internal error"`, which is the actual user-facing interface.
Also in the `Diagnostics` it's marked as `internal compiler error: $Message` when using `bug()`.

197g: Right. I'm changing it to `InternalCompilerError` as proposed.

oberien:
Could you please rebase this PR to the latest changes of error-reporting?

oberien (approved):

PR #91, opened 2019-07-12
Subdirectory reference

### LINKED ISSUE #89: Resolve local links in file's workdir but also from project root
Currently if I have an image in `images/image.png` and want to include it in `chapters/chapter.md` as `[include images/image.png]`, it isn't resolved correctly, because the resolver currently only resolves in the context of the file, which in this case is starting from the directory `chapters/`. Our `to_path` implementation ignores any backreferences, so `../images/image.png` also can't be used.

We should resolve local includes first from the current context's path, and then, if it wasn't found, from the project's working directory.

197g:
Closes: #89

### DIFF
--- a/src/backend/latex/document/thesis.rs
+++ b/src/backend/latex/document/thesis.rs
@@ -12,7 +12,7 @@ use crate::config::Config;
 use crate::diagnostics::Input;
 use crate::error::FatalResult;
 use crate::generator::Generator;
-use crate::resolve::{Context, LocalRelative};
+use crate::resolve::Context;
 
 #[derive(Debug)]
 pub struct Thesis;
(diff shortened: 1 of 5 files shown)


oberien:
Having read the PR, I'm not sure if this is the right approach. In fact, I still think that the whole resolve system is overcomplicated and could be simplified. Honestly I really have trouble not only understanding the changes of this PR, but the resolver in general.

I'm going to go ahead and attempt a complete rewrite in hope to simplify things. I'll also try to add a `resolve_file` method to get rid of `config::resolve_file`. I'll leave this PR open for now in case my rewrite fails.

oberien:
Doesn't look perfect to me, but looks good enough for it to be merged, because I'm currently rewriting a lot of the resolver and getting rid of `Source` in [rewrite-resolve](https://github.com/oberien/heradoc/tree/rewrite-resolve). I like the idea of using URLs instead of Paths for joining purposes.

PR #92, opened 2019-08-05
Rewrite resolve

oberien:


197g:
ETA of review: tomorrow.

src/backend/latex/document/thesis.rs
```diff
@@ -127,17 +127,27 @@ impl<'a> Backend<'a> for Thesis {
...
     let arena = Arena::new();
+    let stderr = Arc::clone(diagnostics.stderr());
     let mut gen = Generator::new(cfg, Thesis, out, &arena, stderr);
     let markdown = fs::read_to_string(&path)?;
-    let context = Context::LocalAbsolute(path.clone());
+    let context = match Context::from_path(path.clone()) {
+        Ok(context) => context,
+        Err(e) => {
+            diagnostics
+                .bug("can't generate context from abstract path")
```
197g: The path here is under user-control. I don't think `bug` is appropriate.

src/generator/mod.rs
```diff
@@ -82,7 +82,18 @@ impl<'a, B: Backend<'a>, W: Write> Generator<'a, B, W> {
     }
 
     pub fn generate(&mut self, markdown: String) -> FatalResult<()> {
-        let context = Context::relative_root(self.cfg.input_dir.clone());
+        // the project root is "." if interpreted as relative to the project root
+        let context = match Context::from_path(".") {
```
197g: Maybe this should simply be an inherent method of `Context`, to create a root context. Then there would be no parsing involved.

src/resolve/include.rs
```diff
@@ -1,9 +1,70 @@
...
+use url::{Url, ParseError};
+
+use super::BASE_URL;
+
+/// Represents the current context we're including from
+#[derive(Debug, PartialEq, Eq)]
+pub struct Context {
+    /// Full URL to the file / content
+    ///
+    /// Can be `heradoc://document/foo/bar.md` (relative) or `file://foo/bar.md` (absolute)
```
197g: Nit: I think it is `file:///foo/bar.md`. Two seperators for the scheme seperation and an additional one since there is no domain part for the `file` scheme.

src/resolve/target.rs
```diff
@@ -0,0 +1,331 @@
...
+                return Err(Error::Diagnostic);
+            },
+        };
+        let inner = match url.scheme() {
+            "heradoc" => match url.domain() {
+                Some("document") => TargetInner::LocalRelative(url.path_segments().unwrap().collect()),
+                Some(domain) => TargetInner::Implementation(domain.to_string()),
+                None => {
+                    diagnostics
+                        .error("no heradoc implementation domain found")
```
197g: That seems unclear to me. We didn't find the specified one, not none at all.
```suggestion
                        .error("heradoc implementation domain not found")
```

src/resolve/target.rs
```diff
@@ -0,0 +1,331 @@
...
+                        .emit();
+                    return Err(Error::Diagnostic);
+                }
+            },
+            "file" => match url.to_file_path() {
+                Ok(path) => TargetInner::LocalAbsolute(path),
+                Err(()) => {
+                    diagnostics.error("error converting url to path")
+                        .with_info_section(range, "defined here")
+                        .error("the file url can't be converted to a path")
```
197g: Seems redundant.

src/resolve/target.rs
```diff
@@ -0,0 +1,331 @@
...
+                }
+            }
+        };
+
+        let inner = match inner {
+            inner @ TargetInner::Implementation(_) => inner,
+            inner @ TargetInner::Remote(_) => inner,
+            TargetInner::LocalAbsolute(abs) => TargetInner::LocalAbsolute(canonicalize(abs)?),
+            TargetInner::LocalRelative(rel) => {
+                assert!(rel.is_relative(), "TargetInner::LocalRelative not relative before canonicalizing: {:?}", rel);
```
197g, line 152: Should this not be a `bug` diagnostics, depending on how sure we are never to change this property?

src/resolve/target.rs
```diff
@@ -0,0 +1,331 @@
...
+            },
+
+            // TODO: discuss proper remote rules
+            // check CORS
+            (ContextType::Remote, TargetInner::Remote(url)) if meta.context.url.domain() == url.domain() => (),
+            (ContextType::Remote, TargetInner::Remote(_)) => {
+                meta.diagnostics
+                    .error("permission denied")
+                    .with_error_section(meta.range, "trying to include this")
+                    .error("CORS request detected")
```
197g: `CORS` hints to me that the resolver is a user-agent properly respecting the [`CORS`](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) http headers, which is not the case. I'd rather stay away from the acronym until we have figured out what to support and how (but it could be used in a note).

```suggestion
                    .error("cross origin request detected")
                    .note("proper CORS is not yet supported")
```

src/resolve/target.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub fn into_include(self, remote: &Remote) -> Result<Include> {
+        let TargetChecked { inner, meta } = self;
+        match inner {
+            TargetInner::Implementation(command) => {
+                match Command::from_str(&command) {
+                    Ok(command) => Ok(Include::Command(command)),
+                    Err(()) => {
+                        meta.diagnostics
+                            .error(format!(
+                                "no heradoc implementation found for domain {:?}",
```
197g: We are mapping domains to target here so I'd rewrite this in terms of the target.

```suggestion
                                "the heradoc domain {:?} does not name a valid command"
```

src/resolve/target.rs
```diff
@@ -0,0 +1,331 @@
...
+                                command
+                            )).with_error_section(meta.range, "defined here")
```
197g: Would it be possible to give suggestions or a link to a guide on existing commands? Open an issue if this is not easily done right now.

src/resolve/target.rs
```diff
@@ -0,0 +1,331 @@
...
+
+/// Guess the type of include based on the file extension.
+///
+/// Used to detect the type of include for relative and absolute file paths or for webrequest
+/// includes that did not receive repsonse with a media type header. Matching is performed purely
+/// based on the file extension.
+fn to_include(
+    path: PathBuf, context: Context, range: SourceRange, diagnostics: &Diagnostics<'_>,
+) -> Result<Include> {
+    // TODO: switch on file header type first
```
197g: No longer `TODO`, above we switch on `content_type` first.

197g (changes requested):

src/resolve/target.rs
oberien: If we change this property, we need to change the lines below as well. This assertion ensures that such a change will be caught during testing. This guarantee is also written in the documentation of `LocalRelative`.

src/resolve/target.rs
oberien: We'll need to write a proper document explaining all features of heradoc similar to the Pandoc User Guide. Once we have that, we can link it here.

oberien:
I fixed your comments and rebased to master. Could you please re-review?

src/resolve/target.rs
197g: You are probably right, this should be caught in the resolver checks if it were to change. Thus assertion instead of `diagnostics.bug()` termination seems fine.

src/backend/latex/document/thesis.rs
```diff
@@ -127,17 +127,28 @@ impl<'a> Backend<'a> for Thesis {
...
     let arena = Arena::new();
+    let stderr = Arc::clone(diagnostics.stderr());
     let mut gen = Generator::new(cfg, Thesis, out, &arena, stderr);
     let markdown = fs::read_to_string(&path)?;
-    let context = Context::LocalAbsolute(path.clone());
+    let context = match Context::from_path(path.clone()) {
+        Ok(context) => context,
+        Err(e) => {
+            diagnostics
+                .error(format!("invalid path to `{}`", abstract_name))
```
197g: This error could still be improved. There is very little context available for the user, as it is missing span information and does not mention that the cause is the resolution of the abstract file. I'm not sure if 'invalid path' is the best wording but can't find any better, it suffices as being reasonably clear with the note.

197g (changes requested):
Good job addressing the other issues.

src/backend/latex/document/thesis.rs
oberien: I hope that I clarified it a little better.

197g (approved):

PR #93, opened 2019-08-07
Stateful codegen

oberien:
Add a stateful codegen unit, which can save state in a Backend, but must be specialized for each Backend.

This is in preparation for https://github.com/oberien/heradoc/pull/15#issuecomment-445473711.


src/backend/mod.rs
```diff
@@ -61,35 +61,35 @@ pub trait Backend<'a>: Debug {
...
+
+    type InlineEmphasis: StatefulCodeGenUnit<'a, Self, ()>;
+    type InlineStrong: StatefulCodeGenUnit<'a, Self, ()>;
+    type InlineStrikethrough: StatefulCodeGenUnit<'a, Self, ()>;
+    type InlineCode: StatefulCodeGenUnit<'a, Self, ()>;
+    type InlineMath: StatefulCodeGenUnit<'a, Self, ()>;
+
+    type Equation: StatefulCodeGenUnit<'a, Self, Equation<'a>>;
+    type NumberedEquation: StatefulCodeGenUnit<'a, Self, Equation<'a>>;
+    type Graphviz: StatefulCodeGenUnit<'a, Self, Graphviz<'a>>;
```
197g, line 94: What is the basis to decide which code gen level to require for a specific item kind?

197g (commented):
Just a question, everything else looks fine.

src/backend/mod.rs
oberien: Direct elements without content / enclosed events are Simple/Medium code gen units. Elements which can contain other events (namely all tags) can intercept the generated output for contained events and thus are `{Stateful,}CodeGenUnits`.

197g: That sounds like a useful distinction that likely deserves a comment in the documentation of the trait.

oberien: Done.

PR #97, opened 2019-08-18
Presentation

oberien:
Largely rewrite beamer from #15.


src/backend/latex/document/beamer.rs
```diff
@@ -0,0 +1,232 @@
...
+        &mut self, level: i32, out: &mut impl Write, range: SourceRange, diagnostics: &Diagnostics<'_>
+    ) -> Result<()> {
+        check_level(level, range, diagnostics)?;
+        let last = self.headings.last().cloned().unwrap_or(0);
+        for level in (last+1)..=level {
+            self.headings.push(level);
+            match level {
+                1 => {
+                    writeln!(out, "\\begin{{frame}}")?;
+                    writeln!(out, "\\Huge\\centering \\insertsection")?;
```
197g, line 59: If I don't want to use a full page sized frame for sections in a presentation (i.e. too short) and don't use *any* section headers then this creates a completely empty page.

src/util.rs
```diff
@@ -0,0 +1,26 @@
+use std::io::{Write, Result};
+use std::fmt;
+
+pub struct OutJoiner<'a, W: Write> {
+    out: W,
+    join_chars: &'a [u8],
```
197g: Based on usage and **`chars`** I would have expected a `&'a str`.

src/backend/latex/document/beamer.rs
```diff
@@ -0,0 +1,232 @@
...
+            write!(out, "{},", other)?;
+        }
+
+        // Beamer already loads internally color, hyperref, xcolor. Correct their options.
+        writeln!(out, "color={{usenames,dvipsnames}},")?;
+        writeln!(out, "xcolor={{usenames,dvipsnames}},")?;
+        writeln!(out, "hyperref={{pdfusetitle}},")?;
+
+        writeln!(out, "]{{beamer}}")?;
+        writeln!(out, "\\usetheme{{Madrid}}")?;
```
197g: Can we include a configuration option for the theme/color? I don't like blue and the solid color bars top and bottom of Madrid :smile:

197g (changes requested):
Much nicer with stateful access to the backend. 

I'm not completely sold on the style for headings but the previous PR was not perfect as well. In all other markdown editors the frame delimiters, the rulers (`---`), appear at a smaller structural level than the boxes and are more easily overlooked.

oberien:
I added a `sectionframes` and `beamertheme` config option.

As said in https://github.com/oberien/heradoc/pull/15#issuecomment-519085285 I followed the default way pandoc handles it.

197g (approved):

PR #106, opened 2019-12-19
acmart latex backend

oberien:
~~Depends on #105~~ It was merged

### DIFF
--- a/src/backend/latex/document/acmarticle.rs
+++ b/src/backend/latex/document/acmarticle.rs
@@ -1,0 +1,92 @@
+use std::io::Write;
+
+use crate::backend::latex::{self, preamble};
+use crate::backend::Backend;
+use crate::config::Config;
+use crate::error::FatalResult;
+use crate::diagnostics::Diagnostics;
+
+#[derive(Debug)]
+pub struct AcmArticle;
+
+#[rustfmt::skip]
+impl<'a> Backend<'a> for AcmArticle {
+    type Text = latex::TextGen;
+    type Latex = latex::LatexGen;
+    type FootnoteReference = latex::FootnoteReferenceGen;
+    type BiberReferences = latex::BiberReferencesGen;
+    type Url = latex::UrlGen;
+    type InterLink = latex::InterLinkGen;
+    type Image = latex::ImageGen;
+    type Label = latex::LabelGen;
+    type Pdf = latex::PdfGen;
+    type SoftBreak = latex::SoftBreakGen;
+    type HardBreak = latex::HardBreakGen;
+    type TaskListMarker = latex::TaskListMarkerGen;
+    type TableOfContents = latex::TableOfContentsGen;
+    type Bibliography = latex::BibliographyGen;
+    type ListOfTables = latex::ListOfTablesGen;
+    type ListOfFigures = latex::ListOfFiguresGen;
+    type ListOfListings = latex::ListOfListingsGen;
+    type Appendix = latex::AppendixGen;
+
+    type Paragraph = latex::ParagraphGen;
+    type Rule = latex::RuleGen;
+    type Header = latex::HeaderGen<'a>;
+    type BlockQuote = latex::BlockQuoteGen;
+    type CodeBlock = latex::CodeBlockGen;
+    type List = latex::ListGen;
+    type Enumerate = latex::EnumerateGen;
+    type Item = latex::ItemGen;
+    type FootnoteDefinition = latex::FootnoteDefinitionGen;
+    type UrlWithContent = latex::UrlWithContentGen<'a>;
+    type InterLinkWithContent = latex::InterLinkWithContentGen;
+    type HtmlBlock = latex::HtmlBlockGen;
+    type Figure = latex::FigureGen<'a>;
+
+    type TableFigure = latex::TableFigureGen<'a>;
+    type Table = latex::TableGen<'a>;
+    type TableHead = latex::TableHeadGen;
+    type TableRow = latex::TableRowGen;
+    type TableCell = latex::TableCellGen;
+
+    type InlineEmphasis = latex::InlineEmphasisGen;
+    type InlineStrong = latex::InlineStrongGen;
+    type InlineStrikethrough = latex::InlineStrikethroughGen;
+    type InlineCode = latex::InlineCodeGen;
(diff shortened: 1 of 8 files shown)


oberien:
I rebased it onto the new master so the changes are easier to review.

src/backend/latex/document/acmarticle.rs
```diff
@@ -0,0 +1,100 @@
...
+        writeln!(out, "\\PassOptionsToPackage{{usenames,dvipsnames}}{{color}}")?;
+        writeln!(out, "\\PassOptionsToPackage{{pdfusetitle}}{{hyperref}}")?;
+        writeln!(out, "\\PassOptionsToPackage{{final}}{{microtype}}")?;
+        preamble::write_documentclass(cfg, out, "acmart", "nonacm,sigconf,natbib=false,pdfusetitle,")?;
+        preamble::write_packages(cfg, out)?;
+        preamble::write_fixes(cfg, out)?;
+
+        if let Some(abstract1) = &cfg.abstract1 {
+            writeln!(out, "\\begin{{abstract}}")?;
+            preamble::gen_abstract(abstract1.clone(), "abstract", &Arena::new(), AcmArticle, cfg, out, diagnostics)?;
```
197g, line 80: I find it weird that this creates a new `AcmArticle` and `Arena` instead of using `&mut self` and shared state. The former seems easier to address but for the latter the generator holding the usual Arena does not yet exist. It's also not a large problem.

src/backend/latex/document/acmarticle.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        writeln!(out)?;
+        writeln!(out, "\\begin{{document}}")?;
+        writeln!(out)?;
+
+        if cfg.title.is_some() {
+            // TODO: Warn if title isn't set but something else is
+            preamble::write_maketitle_info(cfg, out)?;
+            writeln!(out, "\\maketitle")?;
+        }
```
197g, line 92: `write_maketitle_info` internally only writes information that exists. This makes the check appear brittle, i.e. intentionally too lenient. I'd expect the provider of the title page content to have some method of the form  

> `has_titlepage_info(&Config) -> Result<bool>`

which can at least differentiate three cases:
* all info was provided (`Ok(true)`)
* no info was provided and no title page is to be generated (`Ok(false)`)
* partial info was provided but some necessary data was missing (`Err(_)`)

For backends that use the defines in `preamble` or known `acm` standards this could be a common method.

197g: I'd be fine with a method that is not completely comprehensive but this would emphasize how document classes should check and make it easier to refactor and consolidate them.

197g (changes requested):
With the increasing number of backends we should aim to consolidate most configuration dependent logic to avoid cases where changing the document type results in surprising errors.

src/backend/latex/document/acmarticle.rs
oberien: My idea here was that the abstract should be generated in isolation. Right now I don't see a reason why the abstracts should be treated context-sensitively, when they can just be treated context-free. Changing the callsites to pass `&mut self` instead of a new instance is easy enough, but that doesn't change anything as long as the abstracts can be context-free.

Re `Arena` the generator is already created, but reusing its arena means passing the arena to all `gen_preamble` functions. First, most of them don't require the arena (and for me it doesn't make sense to pass a reference to the arena into those functions generator-wise, that would only add more coupling between the generator and backend; also the generator is partially borrowed when calling `gen_preamble`, so we can't just pass a reference to the generator into those functions). Second, again, the abstract doesn't need to know anything about the rest of the document and vice versa. Reusing the global arena would mean that we keep the abstract string stored in memory forever, while creating a new arena will drop the abstract's source as soon as its finished generating.

I can do both of those changes, but I currently don't really see a reason why.

src/backend/latex/document/acmarticle.rs
oberien: You forgot the case that's mentioned in that todo: Info was provided that's not used by maketitle of that documentclass. Not specifying everything should be perfectly fine, I designed the templates in a way that doesn't require all titlepage properties to be set to still work (and maketitle does that as well). Providing partial info should be specifically supported by us, not everyone needs to have the `supervisor` and `advisor` fields for an article, a checklist or something simple. Those fields are only required in some cases and leaving them out shouldn't result in a diagnostic.

In the long term I'd like to have titlepage info be its own struct embedded in the `Config`. When we do that, we should be able to more easily handle all of its fields by destructuring it. That way we won't miss a place it's used when we add another field. That's also when adding those diagnostics should be easy.

src/backend/latex/document/acmarticle.rs
197g: What about the abstract is special that makes it distinct from other sections of the document? In fact, I wonder why the abstract appears in the pre-amble configuration at all. I should be able to cite with full power, put labels or reference them etc. As prior art in customary Latex classes the `abstract` is *contained* in the `document` and not entered before it.

oberien: > I should be able to cite with full power, put labels or reference them etc.

That is a very good point I didn't really think about so far.

> As prior art in customary Latex classes the `abstract` is _contained_ in the `document` and not entered before it.

In our `thesis` (which uses `scrbook`) you're right, that the abstract is part of the `document`. However, it's still part of the `frontmatter` and in the TUM template should be placed before the tableofcontents. That means that we can't move to the `mainmatter` in the `gen_preamble` function.

https://github.com/oberien/heradoc/blob/2c6229f6ce58ab95a35d2d396aca270c6049ac13/src/backend/latex/document/thesis.rs#L100-L114

Additionally, for `acmart`, citing from the specification:

> The environment `abstract` must *precede* the `\maketitle` command. Again, this is different from the standard LATEX.  Putting `abstract` after `\maketitle` will trigger an error.
> – [<https://www.acm.org/binaries/content/assets/publications/consolidated-tex-template/acmart.pdf>, Page 17]

IMO those reasons make it pretty hard to include the `abstract` anywhere else than in the config referring to a different file. We could try some magic forkery where we parse the markdown twice (which we'll need to do anyways at some point due to footnotes), and in the first pass generate the abstract, which could be specified between `[abstract]` and `[endabstract]`.

PR #107, opened 2019-12-20
Add Theorems

### LINKED ISSUE #104: Theorems and proofs
I'd like to add definitions, theorem, lemmata and their corresponding proofs in a fashion similar to that of Latex. These are not necessarily only `math` mode content and their interior should be subject to all other transformations done from Markdown, e.g. lists, quotes, tables...

One prototype for syntax is to allow configuration items on quotes as quotes allow nesting of other markdown blocks. The extra nesting at the start of a paragraph is not quite optimal for long proofs but other options are much worse. 

* Code fences do not parse and allow any inner elements.
* Headings do not have any explicit end markers, see previous attempts with beamer.
* Tables, includes, and links do not add anything obvious to the table
* Lists also have nesting but the semantics seem weird and it is not obvious how multiple items should be interpreted. Other than that they are the same as quotes.

Ideally, this would work and lazy paragraph continuation could remove *some* quote indentation.

````
{theorem, caption="The main theorem"}
> There is an infinite number of prime numbers.

{proof}
> Assume `$ p` to be the largest prime number. Define
> ```$$
P' = 1+ \prod^p_{i=0, i prime} i
```
> Then `$ P'` is not divisible by any prime. But it has at least two divisors, one and itself, so there must a prime divisor larger than `$ p`. This is a contradiction.
````

### PR DESCRIPTION
**Pull request by 197g**
Introduces support for theorems and proofs with very slight extensions to the syntax. Quotes are allowed to have configuration which previously was not possible. This makes them a useful element for embedding scoped content in which the Markdown semantics should be preserved.

Closes: #104

### CODE REVIEW

File: examples/functionality/theorem.md [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+{definition}
+> A natural number `$ n` is called prime if it has exactly two divisors.
+
+{lemma, title="Euclid"}
+> If `$ p` is prime and divides `$ ab` then it divides `$ a` or `$ b`.
+
+{corollary, title="Prime factorization"}
+> Each number has a canonical decomposition into prime factors.
+
+{theorem, title="The main theorem"}
```
oberien, line 12: Is there a reason you chose `title` over `caption`? I think using `caption` instead of `title` might be more consistent with all other configs.

src/backend/latex/complex/proof.rs
```diff
@@ -0,0 +1,57 @@
...
+pub struct ProofGen {
+    kind: ProofKind,
+}
+
+impl<'a> CodeGenUnit<'a, Proof<'a>> for ProofGen {
+    fn new(
+        _cfg: &'a Config, proof: WithRange<Proof<'a>>,
+        gen: &mut Generator<'a, impl Backend<'a>, impl Write>,
+    ) -> Result<Self> {
+        let WithRange(element, _range) = proof;
```
oberien, line 36: Could you destructure the `Proof` here such that it becomes a hard error if we add a new property to it? I did this everywhere else for that exact reason. That way the compiler tells us wherever we need to change something once an attribute is added to something.

src/backend/latex/complex/proof.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl ContextName for ProofKind {
+    fn context_name(&self) -> &'static str {
+        match self {
+            ProofKind::Corollary => "amsthm-corollary",
+            ProofKind::Definition => "amsthm-definition",
+            ProofKind::Lemma => "amsthm-lemma",
+            ProofKind::Proof => "proof",
+            ProofKind::Theorem=> "amsthm-theorem",
```
oberien, line 21:
```suggestion
            ProofKind::Theorem => "amsthm-theorem",
```

src/frontend/mod.rs
```diff
@@ -647,6 +658,45 @@ impl<'a> Frontend<'a> {
...
+            }
+        } else {
+            self.diagnostics
+                .error("quote environment has no type")
+                .with_error_section(range, "configuration defined here")
+                .help("try indicating the type, the configuration could look like")
+                .help("  {proof}")
+                .emit();
+            return;
+        };
```
oberien, line 690: What if I want to have a quote within a figure with a caption, or a quote with a label? We don't support that right now, but with a change like this, I think we should support it in the long run. If you don't want to change it right now, could you please open an issue?

oberien (changes requested):
See comments.

Also the following crashes on this branch:

```md
{figure,#quote,caption=A Quote}
> Fancy Quote
```

oberien:
In general I'd still prefer using the already established code-blocks for this, but as you said, it doesn't have an existing infrastructure surrounding using markdown within the code-block. If we use code-blocks instead of quotes for this, we should be able to allow markdown-parsing by feeding the code-block-content back into pulldown-cmark. I think this might be the better solution syntax-wise, but might result in a harder infrastructure.

examples/functionality/theorem.md
197g: It is not a caption in the sense of a `\\caption{}` element in Latex. It's more closely a heading. It still might be the best fit for consistency.

197g:
I don't think it would be possible with a change short of forking pulldown_cmark. It would require a feedback mode where it is possible to insert content into the source code, potentially affecting even the *tokenization* stage within the parser. Not to mention all other kinds of events where the parser does look-ahead, for example for link callbacks.

Also, the solution would *only* work in heradoc. It would not work in any other tool, which means no editor support, no refactoring support, etc. This compatibility was one primary motivation behind the other syntax choices—including includes and label/description that's proposed. I'm content with the Configuration extension for features that do not affect Markdown parsing but wouldn't use it beyond that.

PR #109, opened 2020-01-23
latex: map lstlisting language `asm` to `[x86_64]{Assembler}`

oberien:


src/backend/latex/complex/codeblock.rs
```diff
@@ -28,6 +28,11 @@ impl<'a> CodeGenUnit<'a, CodeBlock<'a>> for CodeBlockGen {
             joiner.join(format_args!("caption={{{}}}", caption))?;
         }
         if let Some(WithRange(language, _)) = language {
+            // some fixes for weird latex-specific language naming
+            let language = match language.as_ref() {
+                "asm" => "[x86_64]{Assembler}",
```
197g: Maybe slightly more expressive, `asm-x86` for example. Using `ARMv7` of Risc-V assembly shouldn't feel less important and would benefit from language shorthand as well.

oberien: I think that we should also support `asm` as language identifier directly. For example github directly supports `asm` and defaults it to x86 [source 1](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L328-L332), [source 2](https://github.com/jmm/gfm-lang-ids/wiki/GitHub-Flavored-Markdown-(GFM)-language-IDs). That is also the reason I defaulted `asm` to x86.

source 3 (notice the `in` being highlighted in an ARM comment):

```asm
; comment in x86
mov rax, [rdi + rsi * 8]

@ comment in ARM
str r0, [r1]
```

197g: Github is somewhat discriminatory here but crucially it does not disambiguate on name or file extension alone. Rather, it uses some regex matching on unique registers or instructions as further heuristics https://github.com/github/linguist/blob/master/lib/linguist/heuristics.yml.

oberien: I could argue in favour of supporting `asm-x86`, `x86` and `x86-asm`. However, should we really just make `asm` error?

197g: I would argue it is a similar risk as we had with inline latex in the first place. The idea was cute to just write it into the document but the current systematic approach of explicitely opting-into this with a language tag is more principled and robust, especially with regards to adding more functionality such as html output.

Here it is similar, if the standard document contains `asm` instead of the (similiarly ergonomic) `asm-x86` then this means limited choices for change or broken documents with newer versions. Both seem not worth the tradeoff of 4 measly characters.

oberien: Done. Could you please also take a look at #110 ?

PR #110, opened 2020-01-23
Use includegraphics instead of includepdf in Beamer

oberien:
`\includepdf` is not compatible with `beamer` <https://tex.stackexchange.com/a/11461>. Use `\includegraphics` instead.

### DIFF
--- a/src/backend/latex/document/beamer.rs
+++ b/src/backend/latex/document/beamer.rs
@@ -96,7 +96,7 @@ impl<'a> Backend<'a> for Beamer {
     type InterLink = latex::InterLinkGen;
     type Image = latex::ImageGen;
     type Label = latex::LabelGen;
-    type Pdf = latex::PdfGen;
+    type Pdf = latex::BeamerPdfGen;
     type SoftBreak = latex::SoftBreakGen;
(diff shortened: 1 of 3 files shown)


oberien:
@HeroicKatora Could you review this please? (just a notification in case the original notification got swallowed again :) )

197g (commented):
I can't really say for sure. I'm skeptical of stackoverflow comments that give no reference but it seems somewhat correct. How sure are you that we can't have an include outside of a frame in beamer?

oberien:
I wanted to use this to include SVGs by converting them manually to pdfs and including those pdfs. However, I'm currently working on doing that automatically in heradoc, so I guess this PR isn't required for that reason anymore. Instead, thinking about when you want to include a PDF apart from that, I think that when people use ``, they probably want to include that PDF as slides, so this PR is somewhat counterproductive.

I think this is actually currently a bug and trying to include a pdf in `beamer` will always error while running `pdflatex`. We should produce a diagnostic for now that it's not supported until we properly implement including PDFs by closing the frame, using `\includepdf` and checking that the next command will open a new frame.

Closing this in favour of #111.

Issue #98, opened 2019-09-06
Binary artifact size


197g:
The full binary `heradoc` compiles to `94MB` (!) on my system. `strip` still leaves it at moderate but still large `7.7MB`. `cargo bloat` and `debug = true` in the release profile make it clear that a large portion relates the debug information in the text section.

Investigate:
* Is it possible to get useful crahes/backtraces with less bloat?
* Can the size be remove in another manner (e.g. compressed)?

Issue #10, opened 2018-10-23
Remove temp_dir from cfg


oberien:
#4 introduces the tempdir to the config, which is used by `resolve` to store downloaded resources in to avoid the problems of including files from the cache for now (because the tempdir is random, and thus, can't be guessed easily). When #6 is implemented, the `temp_dir` property should be removed from the Config struct in favour of the cache / out_dir properties.

PR #113, opened 2020-01-25
Add svg support, converting it to pdf for latex

oberien:
This was surprisingly hard. I added an `Svg` struct similar to the `Image` struct when encountering an `.svg` file. It supports conversion to a pdf file. This is used by the latex backend to convert svgs into pdf, because svg handling is hard in latex. The svg is converted in rust using `librsvg` and `cairo`, and the resulting pdf is written to the `temp_dir` and then included via `\includegraphics`.

There was a dependency version conflict for `rand`, so I ran a `cargo update`, resulting in the large numbers of lines changed in `Cargo.lock`.

I also implemented a `Size` struct, which we should use everywhere where a user can provide a size, like `width` and `height` values. However, that's for another day and PR, for which I opened #112.

### DIFF
--- a/src/backend/latex/simple.rs
+++ b/src/backend/latex/simple.rs
@@ -203,7 +203,7 @@ pub struct SvgGen;
 
 impl<'a> MediumCodeGenUnit<Svg<'a>> for SvgGen {
     fn gen<'b, 'c>(svg: WithRange<Svg<'a>>, config: &Config, stack: &mut Stack<'b, 'c, impl Backend<'b>, impl Write>) -> Result<()> {
-        let pdf_path = match svg.0.to_pdf_path(&config.temp_dir) {
+        let pdf_path = match svg.0.to_pdf_path(&config.out_dir) {
             Ok(path) => path,
             Err(e) => {
                 stack.diagnostics().error("can't convert svg to pdf")
--- a/src/backend/latex/simple.rs
+++ b/src/backend/latex/simple.rs
@@ -1,15 +1,18 @@
 use std::borrow::Cow;
 use std::io::Write;
+use std::path::Path;
 
 use super::replace::replace;
 use crate::backend::latex::InlineEnvironment;
 use crate::backend::{Backend, MediumCodeGenUnit, SimpleCodeGenUnit};
-use crate::error::Result;
+use crate::config::Config;
+use crate::error::{Result, Error};
 use crate::frontend::range::WithRange;
 use crate::generator::event::{
     BiberReference,
     FootnoteReference,
     Image,
+    Svg,
     InterLink,
     Pdf,
     TaskListMarker,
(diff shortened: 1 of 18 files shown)


src/backend/latex/simple.rs
```diff
@@ -147,39 +150,71 @@ pub struct ImageGen;
...
+
+    inline_fig.write_end(out)?;
+    Ok(())
+}
+
+#[derive(Debug)]
+pub struct SvgGen;
+
+impl<'a> MediumCodeGenUnit<Svg<'a>> for SvgGen {
+    fn gen<'b, 'c>(svg: WithRange<Svg<'a>>, config: &Config, stack: &mut Stack<'b, 'c, impl Backend<'b>, impl Write>) -> Result<()> {
```
197g, line 205: Could the `config` just be accessible through `stack`? Or even make the whole `temp_file` construction something that is unified there. I'm not sure what the best way is to add the error annotations here but I do not think the handling of configuration parameters should be exposed to the backend. It should also not infect so many other parameter lists for a largely unused functionality.

197g (changes requested):
I suppose this didn't improve compile time very much. I do appreciate using Rust for this conversion though.

src/backend/latex/simple.rs
oberien: The Config is already exposed to the backend for (stateful) code gen units:

https://github.com/oberien/heradoc/blob/318d5564f901ba5d62c2f8ef6b4c409f0b6003ef/src/generator/mod.rs#L143-L146

https://github.com/oberien/heradoc/blob/318d5564f901ba5d62c2f8ef6b4c409f0b6003ef/src/generator/code_gen_units.rs#L49-L54

This is for example used by `Graphviz` graph generation:

https://github.com/oberien/heradoc/blob/318d5564f901ba5d62c2f8ef6b4c409f0b6003ef/src/backend/latex/complex/graphviz.rs#L30-L31

Arguably, that is the only usage of `Config` in code gen units I was able to find. The backend *does* need the config during preamble generation. Currently it should be possible to remove the config from code gen units and move output-file generation to `Stack` (and duplicate the method in `Generator`). However, I'd preefer opening an issue with a separate PR as it's not directly necessary for this PR.

(I changed svg to use `out_dir` instead of `temp_dir`, but we want to remove `temp_dir` anyways somewhere in the future #10)

oberien:
The inital compile time is quite high unfortunately. We're up to 387 dependencies from 209… We could probably reduce that number by not using default-features for some packages, which is directly related to #98.

That said, for now I personally put features over compile-time. Refactoring and stuff like that is definitely required at some point, at which we can also try reducing possibly unneeded dependencies (and update other ones to newer breaking versions).

197g (approved):
Seems fine from a feature perspective.

oberien:
Thanks for the review. I opened #114 for removing `Config` from the code gen units.

PR #115, opened 2020-02-05
remove invalid concat string range check

### LINKED ISSUE #76: Investigate Concat
Concat feels somewhat off, especially the assertion about the source ranges. Needs investigation.

### LINKED ISSUE #108: Assertion failure on inline code within quote
Reproduction:

```
> Apparently, `1 +
> 1` equals crash.
```

> thread 'main' panicked at 'assertion failed: range.end == r.start || range.end + 1 == r.start', src/frontend/concat.rs:29:13

oberien:
Fix #108
Fix #76

### DIFF
--- a/src/frontend/concat.rs
+++ b/src/frontend/concat.rs
@@ -32,6 +32,7 @@ impl<'a> Iterator for Concat<'a> {
             // > foo
             // > bar
             // ```
+            assert!(range.end <= r.start);
             range.end = r.end;
 
             let next = match evt {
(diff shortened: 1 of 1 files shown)


src/frontend/concat.rs
```diff
@@ -25,8 +25,13 @@ impl<'a> Iterator for Concat<'a> {
...
-            // TODO: why are both variants needed?
-            assert!(range.end == r.start || range.end + 1 == r.start);
+            // We can't assume that two text events follow each other directly here.
+            // For example a quote results in two text events following each other having different
+            // offsets in the original markdown content string:
+            // ```md
+            // > foo
+            // > bar
+            // ```
             range.end = r.end;
```
197g, line 36: Given this line, we should still at least assert `range.end <= r.start`, or modify the range to the `min, max` of the endpoints of both ranges.

oberien: There isn't a `min` or `max`, because in theory you could have a million nested quotes:

> > > > > > Nested Quote

oberien: I added the `end <= start` assertion.

197g (approved):

PR #116, opened 2020-02-13
Forward all unknown attributes to latex lstlistings

oberien:
I used this for a presentation to specify fontsize (`basicstyle=\tiny`)
for listings. Shouldn't be a long-term solution in general, so unsure if
we actually want to merge this?

### DIFF
--- a/src/cskvp.rs
+++ b/src/cskvp.rs
@@ -160,6 +160,10 @@ impl<'a> Cskvp<'a> {
         self.double.remove(key)
     }
 
+    pub fn take_all_double(&mut self) -> HashMap<Cow<'a, str>, WithRange<Cow<'a, str>>> {
+        self.double.drain().collect()
+    }
+
     /// Removes all elements from `self`.
     ///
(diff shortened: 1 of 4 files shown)


src/cskvp.rs
```diff
@@ -160,6 +160,10 @@ impl<'a> Cskvp<'a> {
         self.double.remove(key)
     }
 
+    pub fn take_all_double(&mut self) -> HashMap<Cow<'a, str>, WithRange<Cow<'a, str>>> {
```
197g, line 163: In `Drop` we warn for all unused attributes of a `Cskvp`. Using this method makes that more or less pointless.

src/frontend/mod.rs
```diff
@@ -411,6 +411,7 @@ impl<'a> Frontend<'a> {
                 } else {
                     Some(WithRange(language, language_range))
                 },
+                attributes: cskvp.take_all_double(),
```
197g, line 463: Maybe we could take some inspiration from html, and have a `style` attribute instead that contains options that should be forwarded? I'm not completely sure either.

PR #117, opened 2020-04-01
Add an espeak+video backend

197g:
Depends on external programs `ffmpeg`, `sox` and `espeak-ng`. Example in `examples/mp4.md`.

### DIFF
--- a/src/backend/ffmpeg/mod.rs
+++ b/src/backend/ffmpeg/mod.rs
@@ -1,0 +1,146 @@
+use std::borrow::Cow;
+use std::io::{self, Write};
+
+use crate::backend::latex::{self, Beamer};
+use crate::backend::{Backend, CodeGenUnit, StatefulCodeGenUnit};
+use crate::config::Config;
+use crate::diagnostics::Diagnostics;
+use crate::error::{FatalResult, Result};
+use crate::frontend::range::{WithRange, SourceRange};
+use crate::generator::event::{Event, Header};
+use crate::generator::Generator;
+
+#[derive(Debug)]
+pub struct SlidesFfmpegEspeak {
+    slides: Beamer,
+}
+
+#[rustfmt::skip]
+impl<'a> Backend<'a> for SlidesFfmpegEspeak {
+    type Text = <Beamer as Backend<'a>>::Text;
+    type Latex = <Beamer as Backend<'a>>::Latex;
+    type FootnoteReference = <Beamer as Backend<'a>>::FootnoteReference;
+    type BiberReferences = <Beamer as Backend<'a>>::BiberReferences;
+    type Url = <Beamer as Backend<'a>>::Url;
+    type InterLink = <Beamer as Backend<'a>>::InterLink;
+    type Image = <Beamer as Backend<'a>>::Image;
+    type Svg = <Beamer as Backend<'a>>::Svg;
+    type Label = <Beamer as Backend<'a>>::Label;
+    type Pdf = <Beamer as Backend<'a>>::Pdf;
+    type SoftBreak = <Beamer as Backend<'a>>::SoftBreak;
+    type HardBreak = <Beamer as Backend<'a>>::HardBreak;
+    type TaskListMarker = <Beamer as Backend<'a>>::TaskListMarker;
+    type TableOfContents = <Beamer as Backend<'a>>::TableOfContents;
+    type Bibliography = <Beamer as Backend<'a>>::Bibliography;
+    type ListOfTables = <Beamer as Backend<'a>>::ListOfTables;
+    type ListOfFigures = <Beamer as Backend<'a>>::ListOfFigures;
+    type ListOfListings = <Beamer as Backend<'a>>::ListOfListings;
+    type Appendix = <Beamer as Backend<'a>>::Appendix;
+
+    type Paragraph = <Beamer as Backend<'a>>::Paragraph;
+    type Rule = PseudoBeamerRuleGen<'a>;
+    type Header = PseudoBeamerHeaderGen<'a>;
+    type BlockQuote = <Beamer as Backend<'a>>::BlockQuote;
+    type CodeBlock = <Beamer as Backend<'a>>::CodeBlock;
+    type List = <Beamer as Backend<'a>>::List;
+    type Enumerate = <Beamer as Backend<'a>>::Enumerate;
+    type Item = <Beamer as Backend<'a>>::Item;
+    type FootnoteDefinition = <Beamer as Backend<'a>>::FootnoteDefinition;
+    type UrlWithContent = <Beamer as Backend<'a>>::UrlWithContent;
+    type InterLinkWithContent = <Beamer as Backend<'a>>::InterLinkWithContent;
+    type HtmlBlock = <Beamer as Backend<'a>>::HtmlBlock;
+    type Figure = <Beamer as Backend<'a>>::Figure;
+
+    type TableFigure = <Beamer as Backend<'a>>::TableFigure;
+    type Table = <Beamer as Backend<'a>>::Table;
+    type TableHead = <Beamer as Backend<'a>>::TableHead;
+    type TableRow = <Beamer as Backend<'a>>::TableRow;
+    type TableCell = <Beamer as Backend<'a>>::TableCell;
+
+    type InlineEmphasis = <Beamer as Backend<'a>>::InlineEmphasis;
+    type InlineStrong = <Beamer as Backend<'a>>::InlineStrong;
+    type InlineStrikethrough = <Beamer as Backend<'a>>::InlineStrikethrough;
+    type InlineCode = <Beamer as Backend<'a>>::InlineCode;
+    type InlineMath = <Beamer as Backend<'a>>::InlineMath;
+
+    type Equation = <Beamer as Backend<'a>>::Equation;
+    type NumberedEquation = <Beamer as Backend<'a>>::NumberedEquation;
+    type Graphviz = <Beamer as Backend<'a>>::Graphviz;
+
+    fn new() -> Self {
+        SlidesFfmpegEspeak {
+            slides: Beamer::new(),
+        }
+    }
+
+    fn gen_preamble(&mut self, cfg: &Config, mut out: &mut impl Write, diagnostics: &Diagnostics<'a>) -> FatalResult<()> {
+        self.slides.gen_preamble(cfg, out, diagnostics)
+    }
+
+    fn gen_epilogue(&mut self, cfg: &Config, out: &mut impl Write, diagnostics: &Diagnostics<'a>) -> FatalResult<()> {
+        self.slides.gen_epilogue(cfg, out, diagnostics)
+    }
+}
+
+#[derive(Debug)]
+pub struct PseudoBeamerRuleGen<'a> {
+    cfg: &'a Config,
+    range: SourceRange,
+}
+
+impl<'a> StatefulCodeGenUnit<'a, SlidesFfmpegEspeak, ()> for PseudoBeamerRuleGen<'a> {
+    fn new(
+        cfg: &'a Config, WithRange(_, range): WithRange<()>,
+        _gen: &mut Generator<'a, SlidesFfmpegEspeak, impl Write>,
+    ) -> Result<Self> {
+        Ok(PseudoBeamerRuleGen { cfg, range })
+    }
+
+    fn finish(
+        self, gen: &mut Generator<'a, SlidesFfmpegEspeak, impl Write>,
+        _peek: Option<WithRange<&Event<'a>>>,
+    ) -> Result<()> {
+        let PseudoBeamerRuleGen { cfg, range } = self;
+        let (diagnostics, backend, mut out) = gen.backend_and_out();
+        backend.slides.close_until(2, &mut out, range, diagnostics)?;
+        backend.slides.open_until(2, cfg, &mut out, range, diagnostics)?;
+        Ok(())
+    }
+}
+
+#[derive(Debug)]
+pub struct PseudoBeamerHeaderGen<'a> {
+    cfg: &'a Config,
+    level: i32,
+    label: WithRange<Cow<'a, str>>,
+    range: SourceRange,
+}
+
+impl<'a> StatefulCodeGenUnit<'a, SlidesFfmpegEspeak, Header<'a>> for PseudoBeamerHeaderGen<'a> {
+    fn new(
+        cfg: &'a Config, header: WithRange<Header<'a>>,
+        gen: &mut Generator<'a, SlidesFfmpegEspeak, impl Write>,
+    ) -> Result<Self> {
+        let (diagnostics, backend, mut out) = gen.backend_and_out();
+        let WithRange(Header { label, level }, range) = header;
+
+        // close old slide / beamerboxesrounded
+        backend.slides.close_until(level, &mut out, range, diagnostics)?;
+
+        write!(out, "\\{}section{{", "sub".repeat(level as usize - 1))?;
+
+        Ok(PseudoBeamerHeaderGen { cfg, level, label, range })
+    }
+
+    fn finish(
+        self, gen: &mut Generator<'a, SlidesFfmpegEspeak, impl Write>,
+        _peek: Option<WithRange<&Event<'a>>>,
+    ) -> Result<()> {
+        let PseudoBeamerHeaderGen { cfg, level, label, range } = self;
+        let (diagnostics, backend, mut out) = gen.backend_and_out();
+        writeln!(out, "}}\\label{{{}}}\n", label.0)?;
+
+        backend.slides.open_until(level, cfg, &mut out, range, diagnostics)?;
+        Ok(())
+    }
+}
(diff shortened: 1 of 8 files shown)


src/backend/ffmpeg/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+        self.slides.gen_preamble(cfg, out, diagnostics)
+    }
+
+    fn gen_epilogue(&mut self, cfg: &Config, out: &mut impl Write, diagnostics: &Diagnostics<'a>) -> FatalResult<()> {
+        self.slides.gen_epilogue(cfg, out, diagnostics)
+    }
+}
+
+impl SlidesFfmpegEspeak {
+    fn open_speech_file(&self, cfg: &Config, diagnostics: &Diagnostics<'_>) -> Result<File> {
```
oberien:
```suggestion
    fn create_speech_file(&self, cfg: &Config, diagnostics: &Diagnostics<'_>) -> Result<File> {
```

src/backend/ffmpeg/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+        let i = self.current_frame.get();
+        let p = cfg.out_dir.join(format!("espeak_{}.txt", i));
+        let res = OpenOptions::new().create(true).write(true).open(&p);
+        match res {
+            Ok(file) => Ok(file),
+            Err(e) => {
+                diagnostics
+                    .error("error creating temporary espeak file for frame")
```
oberien: While for now it is creating a *temporary* espeak file, it most likely won't be in the long run. Instead, it'll just be in the out_dir as everything else should. Would it be possible to rephrase the error message accordingly? Otherwise we can revisit it when actually implementing proper caching, out_dir handling etc.

197g:
Despite having had the right timing to be an April's fool joke, the suggestion is somewhat serious.

oberien:
I really like this PR!!! Even if it was just for April Fool's, I'd have merged it. I'll review it soon-ish™.

Aren't there any TTS engines, which don't sound this robotic and annoying? For example I've seen twitch streamer bots with quite good TTS.

197g:
There are alternatives, many of them backed by cloud infrastructure which I wanted to avoid. Streamers are using a system provided by streamlabs that's ran on their servers and I can't tell you which voice models it uses. I only found this out when looking it up right now but it explains why you aren't able to customize or prefilter messages. There are of course a few possibilities that are fully open source and run locally (please edit the list if you find any other):

* https://github.com/festvox
* https://github.com/CSTR-Edinburgh/merlin

src/backend/latex/document/beamer.rs
```diff
@@ -17,58 +17,83 @@ pub struct Beamer {
...
+    BeginBox,
+    EndFrame,
+    EndBox,
+}
+
 impl Beamer {
     /// Closes the beamerboxesrounded / slides until including the given level.
     /// Also performs the according checks and updates the heading stack.
-    pub fn close_until(
+    pub fn close_until<C: Default + Extend<FrameEvent>>(
```
oberien: Just return a Vec, we shouldn't need anything further than that (for now without optimizations), and it makes everything simpler and more readable.

src/backend/ffmpeg/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+impl CurrentFrame {
+    const fn get(&self) -> u32 {
+        self.0
+    }
+}
+
+impl Extend<BeamerFrameEvent> for CurrentFrame {
+    fn extend<Iter: IntoIterator<Item=BeamerFrameEvent>>(&mut self, iter: Iter) {
```
oberien: While `Extend` might fit functionally, I think it doesn't really imply the intent of this action. What do you think about having a method instead, with a name like `advance_with` or something expressing its function better?

src/backend/ffmpeg/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+        let WithRange(CodeBlock { label: _, caption: _, language }, _range) = &code_block;
+
+        if let Some(WithRange(language, _range)) = language {
+            if language.as_ref() == "espeak" {
+                let (diagnostics, backend, _) = gen.backend_and_out();
+                return backend.open_speech_file(cfg, diagnostics).map(CodeBlockGen::Speech);
+            }
+        }
+
+        <<Beamer as Backend<'static>>::CodeBlock as CodeGenUnit<'a, CodeBlock<'a>>>
```
oberien: 1. No need to specify specific lifetimes here, you can replace both `'static` and `'a` with `'_`.
2. The compiler's inference goes even further, you can just write `CodeGenUnit::new` directly here.

src/backend/ffmpeg/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+            ::new(cfg, code_block, gen)
+            .map(CodeBlockGen::Normal)
+    }
+
+    fn output_redirect(&mut self) -> Option<&mut dyn Write> {
+        match self {
+            CodeBlockGen::Speech(file) => Some(file),
+            CodeBlockGen::Normal(inner) => 
+                <<Beamer as Backend<'static>>::CodeBlock as CodeGenUnit<'a, CodeBlock<'a>>>
```
oberien: ~

src/backend/ffmpeg/mod.rs
```diff
@@ -0,0 +1,241 @@
...
+        }
+    }
+
+    fn finish(
+        self, gen: &mut Generator<'a, SlidesFfmpegEspeak, impl Write>,
+        peek: Option<WithRange<&Event<'a>>>,
+    ) -> Result<()> {
+        match self {
+            CodeBlockGen::Normal(inner) =>
+                <<Beamer as Backend<'static>>::CodeBlock as CodeGenUnit<'a, CodeBlock<'a>>>
```
oberien: ~

src/backend/latex/document/beamer.rs
```diff
@@ -17,58 +17,83 @@ pub struct Beamer {
     headings: Vec<i32>,
 }
 
+pub enum FrameEvent {
```
oberien, line 21: Compiler warnings say that you should derive `Debug` and `Copy` here.

src/backend/latex/document/beamer.rs
```diff
@@ -17,58 +17,83 @@ pub struct Beamer {
...
+            levels.extend(event);
         }
-        Ok(())
+        Ok(levels)
     }
 
     /// Opens the beamerboxesrounded / slides until including the given level, updating the heading
     /// stack.
-    pub fn open_until(
+    pub fn open_until<C: Default + Extend<FrameEvent>>(
```
oberien: ~

src/backend/latex/document/beamer.rs
```diff
@@ -179,3 +204,9 @@ impl<'a> Backend<'a> for Beamer {
         Ok(())
     }
 }
+
+impl Extend<FrameEvent> for () {
+    fn extend<Iter: IntoIterator<Item=FrameEvent>>(&mut self, iter: Iter) {
+        for _ in iter {}
+    }
+}
```
oberien: Not needed if you just use a `Vec`.

src/main.rs
```diff
@@ -128,6 +148,108 @@ fn gen(cfg: &Config, markdown: String, out: impl Write) {
...
 
+fn ffmpeg<P: AsRef<Path>>(pdf: P, cfg: &Config) -> PathBuf {
+    // Start by demuxing the pdf into its frames, which gives us their number.
+    // We use poppler tools and not imagemagick because the latter is bloody stupid. It
+    // disables code access to pdf conversion as a 'security measure'. That's an idiots patch
+    // to broken code that just hurts usability. There's nothing malicious about what we're
+    // attempting to do but, I assumed, since they can't admit to the PHP crowds of SaaS image
+    // tools that their tools are utterly messy and completely unmaintainable they introduce a
+    // black-list that makes it impossible to use for other people as well.
+    // /rant
```
oberien, line 206: :+1:

src/main.rs
```diff
@@ -128,6 +148,108 @@ fn gen(cfg: &Config, markdown: String, out: impl Write) {
...
+    Command::new("pdftoppm")
+        .current_dir(&cfg.out_dir)
+        .arg("-png")
+        .arg(pdf.as_ref())
+        .arg("pages")
+        .status()
+        .expect("Converting pdf with `convert` failed");
```
oberien: Considering the rant above, I'm a bit surprised by this error message :thinking: ;)

src/main.rs
```diff
@@ -128,6 +148,108 @@ fn gen(cfg: &Config, markdown: String, out: impl Write) {
...
+    // disables code access to pdf conversion as a 'security measure'. That's an idiots patch
+    // to broken code that just hurts usability. There's nothing malicious about what we're
+    // attempting to do but, I assumed, since they can't admit to the PHP crowds of SaaS image
+    // tools that their tools are utterly messy and completely unmaintainable they introduce a
+    // black-list that makes it impossible to use for other people as well.
+    // /rant
+    Command::new("pdftoppm")
+        .current_dir(&cfg.out_dir)
+        .arg("-png")
+        .arg(pdf.as_ref())
```
oberien:
```suggestion
        .args(&["-png", pdf.as_ref()])
```
I think grouping args here eases readability.

src/main.rs
```diff
@@ -128,6 +148,108 @@ fn gen(cfg: &Config, markdown: String, out: impl Write) {
...
+    let mut audios = vec![];
+
+    for idx in 0.. {
+        // Yes, for some reason the page index is 1 based
+        let frame = format!("pages-{}.png", idx + 1);
+        let speak = format!("espeak_{}.txt", idx);
+        let wav = format!("espeak-{}.wav", idx);
+
+        if !cfg.out_dir.join(&frame).exists() {
+            break;
```
oberien, line 232: Shouldn't this be an ICE / output an ICE message?

src/main.rs
```diff
@@ -128,6 +148,108 @@ fn gen(cfg: &Config, markdown: String, out: impl Write) {
...
+
+    // Then, generate all voice files and record their lengths. Note that a voice file need not
+    // exist (e.g. for title frames) in which case we should prepare some filler (TODO).
+    // Also add all the pages to readable image list for ffmpeg
+    let mut control = File::create(cfg.out_dir.join("ffmpeg.concat.txt"))
+        .expect("Failed to create ffmpeg control file");
+
+    let mut audios = vec![];
+
+    for idx in 0.. {
```
oberien, line 223: TOCTOU, but should be good enough for our case.

src/main.rs
```diff
@@ -128,6 +148,108 @@ fn gen(cfg: &Config, markdown: String, out: impl Write) {
...
+            writeln!(control, "duration {}", 0.0).unwrap();
+            continue;
+        }
+
+        Command::new("espeak-ng")
+            .current_dir(&cfg.out_dir)
+            .arg("-f")
+            .arg(&speak)
+            .arg("-w")
+            .arg(&wav)
```
oberien:
```suggestion
            .args(&["-f", &speak])
            .args(&["-w", &wav])
```

src/main.rs
```diff
@@ -128,6 +148,108 @@ fn gen(cfg: &Config, markdown: String, out: impl Write) {
...
+            .args(&["-v", "Henrique"])
+            .status()
+            .expect("Conversion with `espeak-ng` failed.");
+
+        let output = Command::new("sox")
```
oberien: If we depend on `ffmpeg` anyway, why don't we use `ffprobe` / the ffmpeg tool suite to get the wav's metadata?

src/main.rs
```diff
@@ -128,6 +148,108 @@ fn gen(cfg: &Config, markdown: String, out: impl Write) {
...
+            .parse()
+            .expect("Length not a valid value.");
+        writeln!(control, "file '{}'", frame).unwrap();
+        writeln!(control, "duration {}", duration).unwrap();
+        audios.push(wav);
+    }
+
+    // concatenate all audio
+    {
+        let mut concat = Command::new("sox");
```
oberien: Again, `ffmpeg` could do this as well.

oberien (commented):
The slides in the output seem to have a pretty bad quality, could we export them with a better quality / in a higher resolution?

I think the error messages when a binary (dependency) doesn't exist could be improved.

NB: What do you think about adding an `espeak` backend in general, which generates a `flac`, `mp3`, … from just normal markdown?

src/main.rs
197g: Don't you see, I'm expecting that converting with `convert` fails :see_no_evil: No, I just missed this message while replacing it.

src/main.rs
197g: This one is intentional. `pdf.as_ref()` is `&Path` while the former is `&str` so we would have to wrap either with more explicit converters which would, imo, negate some of the readability gained from grouping these arguments.

src/main.rs
197g: This was the sloppy implementation for detecting the number of frames :)

src/main.rs
197g: It was non-obvious how to do this with `ffmpeg`, if you have some pointers to the documentation then I'll gladly switch the implementation.

src/main.rs
197g: That's right, and the concat filter is already used as well. That one should be easy to switch.

src/backend/latex/document/beamer.rs
197g: With this style, the compiler defaults this parameter to a simple `()` if it is not provided and the method is used as a `;`-terminated statement. This is due to magic. It was convenient though and means no allocation when the result is unused which was very satisfying as it just happens to click together nicely.

src/backend/ffmpeg/mod.rs
197g: The original intent was to use `FromIterator`. I could rewrite `close_until` to produce an iterator that can be used but it is slightly more involved. The next best trait was `Extend` which is somewhat more powerful but, imo, serves a very similar purpose in the sense that it is mostly transparent to the caller which implementation is being used.

src/backend/ffmpeg/mod.rs
197g: Hm, weird. I remember that inference had failed somewhere. But I'll remove those superficial annotations then.

src/backend/latex/document/beamer.rs
oberien: Dead-code analysis should also get rid of the allocations. And even if not, I think for now clarity is more important than performance, especially when we're talking about a few allocations over the whole lifetime of the program. Our bottleneck is not the rust part, but instead latex. Optimizations can come later.

src/backend/ffmpeg/mod.rs
oberien: I understood that reasoning, having tried to rewrite it as `FromIterator`, which doesn't work easily. Extend definitely is a nice solution generically, but I still think it hurts ease of understanding the code.

src/main.rs
oberien: While it's possible with `ffmpeg`, using `ffprobe` is probably easier (and `ffprobe` is also part of (at least the arch) ffmpeg installs. You can see both solutions [here](https://superuser.com/a/945604).

src/backend/latex/document/beamer.rs
197g: The only kinds of allocations for which I know that llvm's dead code analysis actually can get rid of are `alloca`. Whether it considers others a side effect, or there is some bad interaction going on but it is _very_ hard for it to reason about any other kinds of allocator use. It's annoying.

197g: Actually that is not exactly true. A direct pair of `alloc`/`dealloc` is optimized away but even one call to `realloc` in-between seems to confuse it enough to give up. [See this playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=3f8c116624e0a4ffe385d592f8a743e1).

oberien: Yeah, I remember llvm being weird around `malloc` and `free`, most of the times treating them like black boxes unfortunately. I can definitely see that your approach is pretty fancy and cool and deep dark magic and all, and I also really love those solutions, but I think in this case it doesn't add much (for now) while hurting readability. Premature optimization is the root of all evil ;)

src/main.rs
oberien: Ohhh, I see. Ewwww!!! At least add a TODO :D

src/main.rs
```diff
@@ -128,6 +150,155 @@ fn gen(cfg: &Config, markdown: String, out: impl Write) {
...
+    // We use poppler tools and not imagemagick because the latter is bloody stupid. It
+    // disables code access to pdf conversion as a 'security measure'. That's an idiots patch
+    // to broken code that just hurts usability. There's nothing malicious about what we're
+    // attempting to do but, I assumed, since they can't admit to the PHP crowds of SaaS image
+    // tools that their tools are utterly messy and completely unmaintainable they introduce a
+    // black-list that makes it impossible to use for other people as well.
+    // /rant
+    Command::new("pdftoppm")
+        .current_dir(&cfg.out_dir)
+        .arg("-png")
```
oberien: For better Resolution of the images in the mp4 (currently they only have 100 or 150 dpi, which looks pixelated).

```suggestion
        .args(&["-png", "-rx", "600", "-ry", "600"])
```

oberien (changes requested):
The changes look good to me so far. See comment inline.

PR #121, opened 2020-04-20
Add svgbob

### LINKED ISSUE #119: Add support for svgbob
[svgbob](https://github.com/ivanceras/svgbob) is a library to convert ascii-art to rendered svgs.

In the frontend, if the language of a codeblock is `svgbob`, render it using `svgbob`, save it to a file `svgbob{n}.svg` in the out-dir and return an include to that file (maybe won't work because resolving the file would fail due to permissions?).

oberien:
Fix #119


197g (approved):
The addition to link-resolution looks solid enough for the forseeable future, it remains to be seen if it proves optimal in the long term.

PR #122, opened 2020-07-02
Update svgbob to build on stable again and fix maketitle short author for non-beamer

oberien:


src/backend/latex/preamble.rs
```diff
@@ -121,8 +133,10 @@ pub fn write_maketitle_info(cfg: &Config, out: &mut impl Write) -> Result<()> {
     }
     if cfg.author.is_some() || cfg.supervisor.is_some() || cfg.advisor.is_some() {
         write!(out, "\\author")?;
-        if let Some(author) = &cfg.author {
-            write!(out, "[{}]", author)?;
+        if short_author == ShortAuthor::Yes {
+            if let Some(author) = &cfg.author {
+                write!(out, "[{}]", author)?;
```
197g, line 141: What's the need for writing the exact same author again? Shouldn't this be complemented by another config option for allowing some customization of the short style and defaulting to the original option?

197g (changes requested):

src/backend/latex/preamble.rs
oberien: If I don't provide a short author, IIRC no author is shown in the footer of slides at all.

oberien: I seem to remember wrongly…

oberien: I opened #125 for it.

PR #126, opened 2020-07-02
Add pagebreak

oberien:


197g (approved):

197g:
You smuggled some reworks into this. Not that I'm mad about :smiley:

oberien:
Thanks for the review! Those were actually required. Hrule is a Tag with Start and End, so we were able to use `StatefulCodeGenUnit`. However, PageBreak is not a Tag with Start and End (and I still don't know why hrule is), so I needed to be able to allow `StatefulCodeGenUnit`s for the simple code gen units as well.

PR #127, opened 2020-07-02
Add algorithm to estimate good layout of tables

oberien:
by guessing their perfect column size correctly

### DIFF
--- a/src/frontend/mod.rs
+++ b/src/frontend/mod.rs
@@ -16,6 +16,7 @@ mod event;
 pub mod range;
 mod refs;
 mod size;
+mod table_layout;
 
 pub use self::event::*;
 pub use self::size::*;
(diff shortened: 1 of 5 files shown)


src/frontend/mod.rs
```diff
@@ -718,10 +728,36 @@ impl<'a> Frontend<'a> {
...
+        let mut column_lines = vec![Vec::new(); alignment.len()];
+        loop {
+            match &self.parser.peek().unwrap().0 {
+                &CmarkEvent::Start(CmarkTag::TableHead) | &CmarkEvent::Start(CmarkTag::TableRow) => {
+                    let mut i = 0;
+                    loop {
+                        let cell_text = match &self.parser.peek().unwrap().0 {
+                            &CmarkEvent::Start(CmarkTag::TableCell) => self.concat_until_end_inclusive(false),
+                            &CmarkEvent::End(CmarkTag::TableHead) | &CmarkEvent::End(CmarkTag::TableRow) => break,
+                            _ => unreachable!(),
```
197g: Can you add a comments to these unreachables / some message?
Not necessarily for the current program, I think they are correct, but just in case common mark or our parsing style changes and these turn into bugs.

197g (commented):
I can't say I've fully understood _why_ the column widths are better but it's fine. Integration also looks amenable for changes or another algorithm if necessary.

src/frontend/mod.rs
oberien: Done. Good idea.

197g (approved):

PR #132, opened 2020-07-03
Fix thickhline for ltablex's convertXColumns

oberien:


197g (commented):
I don't speak `tex` well enough to judge the real change here but the cell separator looks okay.

oberien:
Exactly my thoughts ;) I just verified that it works in both cases now (full-width and non-full-width table).

PR #136, opened 2020-07-08
Windows Support

oberien:
Compile on windows, but without svg→png support, because librsvg on windows is very hard to compile. Also provide build instructions for Windows.

I tried getting it to build in WSL, but after over 6 hours of dependency-chasing, and finally getting it to compile, link failed because some libcairo methods couldn't be found.


197g (approved):
Not bad, actually much more contained than I was fearing.

Issue #138, opened 2020-07-13
Proper error reporting for toml config errors


oberien:
Toml has a [`Spanned`](https://docs.rs/toml/0.5.6/toml/struct.Spanned.html) type which reports parsing span information. This should allow us to have proper diagnostics / error reporting when validating the configs.

Depends on #137

Issue #137, opened 2020-07-13
Refactor Codespan-Reporting / Diagnostics


oberien:
The `codespan` and `codespan-reporting` crates been updated quite a lot. We should incorporate those changes into heradoc. Additionally, the diagnostics should be refactored in a way that we can reuse them for error reporting in the toml files.

- [ ] Update codespan and codespan-reporting
- [ ] Refactor Diagnostics to be reused for any file with a span

PR #141, opened 2020-07-16
Implement Project and project-relative resolving

oberien:


src/main.rs
```diff
@@ -79,31 +79,56 @@ fn main() {
...
+            }
+        } else {
+            None
+        }
     };
 
+    let file = match cfgfile_folder.as_ref() {
+        Some(folder) => {
+            let file = folder.join("heradoc.toml");
+            let content = fs::read_to_string(&file).expect(&format!("error reading config file at {:?}",file));
```
197g, line 106: A panic on user input?

src/main.rs
```diff
@@ -79,31 +79,56 @@ fn main() {
...
-        let content = fs::read_to_string(cfgfile).expect("error reading existing config file");
-        toml::from_str(&content).expect("invalid config")
-    } else {
-        FileConfig::default()
+    // try to find a heradoc.toml
+    // 1. passed via commandline
+    // 2. next to the input file
+    // 3. somewhere in the parent folders of the input file
+    let cfgfile_folder = match args.configfile.as_ref() {
+        Some(file) => Some(file.parent().expect("configfile passed as argument should point to a file and not the root-folder").to_owned()),
```
197g, line 87: I see we have enough of these panic-unwraps above as well. Oh well, it might be time for a proper error with variants.

197g (commented):
Resolution and canonicalization seem fine but we should really, really switch to an error for main instead of panicking. Tell me if you're fine with doing it as part of this PR or a separate one.

oberien:
Thanks for the review! Currently, we always panic on user input regarding command line arguments. I created #138 to track moving to proper diagnostics during argument usage and config parsing / migration. However, that depends on #137, which will be quite a huge refactor. I don't want to block this PR on all of that mountain of work, especially because currently we just panic everywhere before we call into the generator.

197g (approved):

PR #155, opened 2020-09-15
Improve windows build description

oberien:


197g (commented):
I never built on Windows soo... It sounds vaguely correct. I'd rather there would be an initial probe phase for all the dependencies and a clean error as well.

oberien:
We should test for required binaries and abort, and maybe also for optional ones, producing a warning. However, that's not what this PR is for, this PR just improves the description about how to properly install the binary dependencies on Windows.

197g (approved):

